PR #10, opened 2020-02-07
add stream GC and wait with progress function

vchuravy:


lcw:
It seems to me the GC for streams is not thread safe.  Is the assumption that for multi-gpu code something like MPI.jl will be used?

vchuravy:
>  Is the assumption that for multi-gpu code something like MPI.jl will be used?

At least for now, we can certainly relax that later, but multi-GPU single host requires a bit more thought.

PR #11, opened 2020-02-08
[WIP] fix localmem and add dynamic mem support

vchuravy:
still fails with illegal memory access in kernel


Moelf:
how far are we from this? I guess another thing is what defines "dynamic", it seems to have no problem in allocatiin a localmem based on `prod(@groupsize))`

PR #12, opened 2020-02-12
Adding a few more examples

leios:
Here are 2 quick examples for matrix multiplication (not element-wise) and a naive transpose. I am working on a shared mem tile version of this as well.


mwarusz:
This kernels could be made simpler by using the existing Cartesian indexing functionality directly. For example, the transpose could be written as
```
@kernel function naive_transpose!(b, a)
  I = @index(Global, Cartesian)
  i, j = Tuple(I)
  @inbounds b[i, j] = a[j, i]
end
```
You probably want to explicitly `wait` for the enqueued kernels to finish. If you want useful timings, it would be good to not measure compilation time and maybe report a more meaningful number like the achieved memory bandwidth.

leios:
@mwarusz Right, in this case that works. I'll update this example and also add some shared mem usage to a non-naive transpose.

examples/transpose.jl
```diff
@@ -0,0 +1,55 @@
...
+end
+
+@kernel function naive_transpose!(a,b, dim1, dim2)
+    cartesian_in = CartesianIndices(a)[@index(Global)]
+
+    # Changing index order for transpose.
+    # Note: the only way I can see to do this is to transfer it into an array
+    #       for a swap, and then transfer back
+    tmp_idx = MVector(cartesian_in.I)
+    @inbounds tmp_idx[dim1], tmp_idx[dim2] = tmp_idx[dim2], tmp_idx[dim1]
```
vchuravy: We probably should provide a general method for doing `permutedims` on indexes.

examples/transpose.jl
```diff
@@ -0,0 +1,55 @@
...
+    CuArrays.@time naive_transpose!(CUDA(),256)(d_a, d_b, 1, 2, ndrange=size(a))
+
+    a = Array(d_a)
+    b = Array(d_b)
+
+    if (a == transpose(b))
+    #if (a == b)
+        println("Good job, man")
+    else
+        println("You failed. Sorry.")
```
vchuravy: use `@test` here

examples/transpose.jl
leios: I think @mwarusz was right. This is an example. Let's keep it simple. No reason to do multi-dimensional distributed stuff.

leios:
Ok, I made the examples more CPU / GPU generic; however, I cannot quite figure out why the matrix multiplication fails on the CPU. The test does pass on the GPU.

Also, what architecture is available for use right now? is is just `CUDA()` and `ThreadedCPU()`?

examples/matmul.jl
```diff
@@ -0,0 +1,47 @@
+using KernelAbstractions, CuArrays, Test, CUDAapi
```
vchuravy: The load of `CuArrays` needs to be dependent on `CUDAapi.has_cuda_gpu()`

examples/matmul.jl
```diff
@@ -0,0 +1,47 @@
...
+
+# Simple kernel for matrix multiplication
+@kernel function matmul!(a, b, c)
+    if size(a)[2] != size(b)[1]
```
vchuravy: Do the verification in a wrapper function

examples/matmul.jl
```diff
@@ -0,0 +1,47 @@
...
+        # here, we need a CPU / GPU generic print statement, like...
+        # CUDAnative.@cuprintf("Matrix size mismatch!")
+        return nothing
+    end
+    cI = CartesianIndices(c)[@index(Global)]
```
vchuravy:
```suggestion
    cI = @index(Global, Cartesian)
```

examples/matmul.jl
```diff
@@ -0,0 +1,47 @@
...
+    c[cI] = tmp
+end
+
+function check()
+    a = rand(256,123)
+    b = rand(123, 45)
+    c = zeros(256, 45)
+
+    # beginning CPU tests
+    matmul!(ThreadedCPU(),4)(a, b, c, ndrange=size(c))
```
vchuravy:
```suggestion
    ev = matmul!(ThreadedCPU(),4)(a, b, c, ndrange=size(c))
    wait(ev)
```

vchuravy:
bors try

vchuravy:
bors try

leios:
Should I add a shmem example to this PR or create a new one?

vchuravy:
> Should I add a shmem example to this PR or create a new one?

Create a new one.

examples/matmul.jl
```diff
@@ -0,0 +1,66 @@
...
+    println("Testing CPU matrix multiplication...")
+    @test isapprox(a*b, c)
+
+    # beginning GPU tests
+    if has_cuda_gpu()
+        d_a = CuArray(a)
+        d_b = CuArray(b)
+        d_c = CuArray(c)
+
+        launch_matmul!(d_a, d_b, d_c)
```
vchuravy:
```suggestion
        ev = launch_matmul!(d_a, d_b, d_c)
        wait(ev)
```

PR #41, opened 2020-02-28
Allow CPU and CUDA kernels to wait on each other

lcw:
This allows the event system to be used to coordinate work done on the
host with work done on the device.

### DIFF
--- a/src/backends/cpu.jl
+++ b/src/backends/cpu.jl
@@ -41,7 +41,18 @@ end
 function __run(obj, ndrange, iterspace, args, dependencies)
     return Threads.@spawn begin
         if dependencies !== nothing
-            Base.sync_end(map(e->e.task, dependencies))
+            cpu_tasks = Core.Task[]
+            for event in dependencies
+                if event isa CPUEvent
+                    push!(cpu_tasks, event.task)
+                end
+            end
+            !isempty(cpu_tasks) && Base.sync_end(cpu_tasks)
+            for event in dependencies
+                if !(event isa CPUEvent)
+                    wait(event, ()->yield())
+                end
+            end
         end
         @sync begin
             # TODO: how do we use the information that the iteration space maps perfectly to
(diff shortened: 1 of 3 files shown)


lcw:
bors try

src/backends/cpu.jl
```diff
@@ -41,7 +41,15 @@ end
 function __run(obj, ndrange, iterspace, args, dependencies)
     return Threads.@spawn begin
         if dependencies !== nothing
-            Base.sync_end(map(e->e.task, dependencies))
+            cpu_tasks = Core.Task[]
+            for event in dependencies
+                if event isa CPUEvent
+                    push!(cpu_tasks, event.task)
+                else
+                    wait(event)
```
vchuravy:
```suggestion
                    wait(event, ()->yield())
```

I am worried about busy-waiting on the GPU causing us the block a thread from making progress. So we should pass a CPU yield as remedy.

src/backends/cpu.jl
```diff
@@ -41,7 +41,15 @@ end
 function __run(obj, ndrange, iterspace, args, dependencies)
     return Threads.@spawn begin
         if dependencies !== nothing
-            Base.sync_end(map(e->e.task, dependencies))
+            cpu_tasks = Core.Task[]
+            for event in dependencies
```
vchuravy: In general I would invert this scheme to first wait on a CPU events, and then wait on the GPU events.
The CPU wait is more efficiently implemented in the scheduler and we don't have to busy-wait on the GPU.

src/backends/cpu.jl
lcw: Done, and added a yield in the cuda backend as well.

lcw:
bors try

src/backends/cpu.jl
lcw: Okay, I did this and rearranged the cuda backed to first enqueue the gpu waits and then wait on the CPU.

vchuravy:
bors try

PR #47, opened 2020-03-04
[WIP] Infrastructure to sync CuDefaultStream()

### LINKED ISSUE #46: How best sync with the default stream in the CUDA backed?
When combining KA with CuArrays we have needed to sync with the default stream.  The hack we have right now can be seen for example here <https://github.com/climate-machine/CLIMA/pull/799/commits/417dd726d4340e7688f3810b981077dcdadf92bf>.

Should a mechanism exist in KA to give us an event in the default stream to pass as a dependency to KA kernels?

If so what should the CPU backend do?  We could return `nothing` and make `wait(::Nothing; progress=nothing) = nothing`.

vchuravy:
fixes #46

### DIFF
--- a/src/KernelAbstractions.jl
+++ b/src/KernelAbstractions.jl
@@ -2,7 +2,7 @@ module KernelAbstractions
 
 export @kernel
 export @Const, @localmem, @private, @uniform, @synchronize, @index, groupsize
-export Device, GPU, CPU, CUDA 
+export Device, GPU, CPU, CUDA, Event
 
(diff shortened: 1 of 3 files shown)


lcw:
This looks like what I need!

vchuravy:
```
ev = Event(CUDA())
# do work on the default stream
ev = kernel(..., dependencies(ev,))
wait(CUDA(), ev)
# do work on the default stream
```

PR #48, opened 2020-03-04
WIP: Asynchronous copy

### LINKED ISSUE #40: Integrate `async_copy!` into the event system
The code from https://github.com/climate-machine/CLIMA/blob/ad0508d7c0fa1004552197bb32334cfe923dd6ba/src/Arrays/CMBuffers.jl#L124-L209 needs to be integrated into KA and especially into the KA event system.

leios:
This is a WIP PR for asynchronous copies between CPU and GPU streams

For the CPU variant, I am literally just calling an `unsafe_copyto!(...)`.

I need to properly use pinned memory for the GPU->GPU case as well. Working on it.

This fixes #40 (hopefully)


src/KernelAbstractions.jl
```diff
@@ -1,14 +1,19 @@
 module KernelAbstractions
 
 export @kernel
-export @Const, @localmem, @private, @uniform, @synchronize, @index, groupsize
+export @Const, @localmem, @private, @uniform, @synchronize, @index, groupsize, pin!, recordevent, async_copy!
```
vchuravy:
```suggestion
export @Const, @localmem, @private, @uniform, @synchronize, @index, groupsize,
export pin!, recordevent, async_copy!
```

src/KernelAbstractions.jl
```diff
@@ -1,14 +1,19 @@
...
 export Device, GPU, CPU, CUDA 
 
 using MacroTools
 using StaticArrays
 using Cassette
 using Requires
 using Adapt
+using CUDAdrv
```
vchuravy: Any code dependent on CUDAdrv being available needs to move into `backends/cuda.jl`

src/KernelAbstractions.jl
```diff
@@ -62,10 +67,63 @@ abstract type Event end
 import Base.wait
 
 # TODO
-function async_copy! end
 # function register end
 # function unregister end
 
+# Functions for async_copy
+function pin!(a)
```
vchuravy: probably want a `pin!(::CUDA, a)` so that we can pin with respect to different backends.

src/KernelAbstractions.jl
```diff
@@ -62,10 +67,63 @@ abstract type Event end
...
+  ad = Mem.register(Mem.Host, pointer(a), sizeof(a))
+  finalizer(_ -> Mem.unregister(ad), a)
+end
+
+function recordevent(stream)
```
vchuravy: See  #47 

I will try to get that done and then you can rebase.

src/KernelAbstractions.jl
```diff
@@ -62,10 +67,63 @@ abstract type Event end
...
+  event = CuEvent(CUDAdrv.EVENT_DISABLE_TIMING)
+  CUDAdrv.record(event, stream)
+  return CudaEvent(event)
+end
+
+function async_copy!(A,B)
+    unsafe_copyto!(pointer(A), pointer(B), length(A))
```
vchuravy: I am thinking about the syntax here, but I feel like we need
to say something like: `async_copy!(::CPU, A, B, dependencies=nothing)` and actually launch a task that waits on the dependencies before copying. Also we probably should just call `copyto!`.

src/KernelAbstractions.jl
```diff
@@ -62,10 +67,63 @@ abstract type Event end
...
+end
+
+if has_cuda_gpu()
+    function async_copy!(A::CuArray, B::CuArray; stream=CuDefaultStream(),
```
vchuravy: The stream chosen here should come from the `stream pool` in `backends/cuda.jl`.

also probably means we want `async_copy!(::CUDA, )`

vchuravy:
Superseded by #53 thank you James.

PR #52, opened 2020-03-06
fix function name check

GiggleLiu:
Skipped a check that gives error for non-symbolic function names.
Not sure this fix is proper, but it works for my application.


vchuravy:
bors rty

vchuravy:
bors try

vchuravy:
Can you add a test that makes sure that one can overload such kernels?

GiggleLiu:
> Can you add a test that makes sure that one can overload such kernels?

Yes, there is such a test.

test/test.jl
```diff
@@ -216,3 +216,15 @@ if has_cuda_gpu()
     @test MultiEvent((event1, event2, event3)) isa Event
   end
 end
+
+  @testset "fallback test: callable types" begin
+    function f end
+    @kernel function (a::typeof(f))(x, ::Val{m}) where m
```
vchuravy: Can you add a second test:

```
     @kernel function (a::typeof(f))(x, ::Val{1})
```

to see what happens when two invocations of `@kernel` define the wrappers? You should get a warning from Julia

GiggleLiu: I added a test, everything so far works fine? But I didn't get a warning.

GiggleLiu:
Hi, is it good to merge, or are there some further suggestions?

PR #53, opened 2020-03-06
Implement async_copy!

vchuravy:
bors r+

### DIFF
--- a/src/KernelAbstractions.jl
+++ b/src/KernelAbstractions.jl
@@ -3,6 +3,8 @@ module KernelAbstractions
 export @kernel
 export @Const, @localmem, @private, @uniform, @synchronize, @index, groupsize
 export Device, GPU, CPU, CUDA, Event
+export async_copy!
+
 
 using MacroTools
 using StaticArrays
@@ -61,10 +63,12 @@ macro Const end
 abstract type Event end
 import Base.wait
 
-# TODO
+"""
+    async_copy!(::Device, dest::AbstractArray, src::AbstractArray; dependencies = nothing)
+
+Perform an asynchronous copy on the device. Returns an event that can be waited upon.
+"""
 function async_copy! end
-# function register end
-# function unregister end
 
 ###
 # Kernel language
--- a/src/backends/cpu.jl
+++ b/src/backends/cpu.jl
@@ -15,9 +15,28 @@ function wait(::CPU, ev::CPUEvent, progress=nothing)
     else
         while !Base.istaskdone(ev.task)
             progress()
-            yield() # yield to the scheduler
         end
     end
+end
+function __waitall(::CPU, dependencies, progress)
+    if dependencies isa Event
+        dependencies = (dependencies,)
+    end
+    if dependencies !== nothing
+        dependencies = collect(dependencies)
+        cpudeps   = filter(d->d isa CPUEvent && d.task !== nothing, dependencies)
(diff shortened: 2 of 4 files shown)


src/backends/cuda.jl
```diff
@@ -47,39 +47,60 @@ end
...
-# do a CPU wait, and therefore block anyone from submitting more work.
-# We maybe could do a spinning wait on the GPU and atomic flag to signal from the CPU,
-# but which stream would we target?
-wait(::CUDA, ev::CPUEvent,  progress=nothing) = wait(CPU(), ev, progress)
+    return CudaEvent(event)
+end
 
-function __enqueue_wait(ev::CudaEvent, stream::CuStream)
-    CUDAdrv.wait(ev.event, stream)
+function pin!(::CUDA, a)
```
leios: Do the users actually need to see the `pin!(...)` function? I feel like we should just do it for the GPU-> GPU copy and not show anyone. This function is naturally inefficient on the CPU, so having the user allocate unused memory seems a bit weird.

lcw: Yes, I think users need to see `pin!`.  Some users may want to manage their own data transfers between the host and device.  Also, I don't understand why this method is inefficient on the CPU.

src/backends/cpu.jl
```diff
@@ -15,10 +15,33 @@ function wait(::CPU, ev::CPUEvent, progress=nothing)
...
         end
     end
 end
+function __waitall(::CPU, dependencies, progress)
+    if dependencies isa Event
+        dependencies = (dependencies,)
+    end
+    if dependencies !== nothing
+        dependencies = collect(dependencies)
+        cpudeps  = filter(d->d isa CPUEvent,    dependencies)
```
lcw:
```suggestion
         cpudeps = filter(d->d isa CPUEvent && d.task isa Core.Task, dependencies)
```

src/backends/cuda.jl
vchuravy: The issue is that `pin!` is costly, but we could cache the pointer...

leios: I just noticed there is no `pin!(::CPU, a)` function in this PR right now, so I guess if it's a GPU-specific function, that's fine. I didn't want users to allocate a new array to pin in the sake of language agnostic code.

src/KernelAbstractions.jl
```diff
@@ -3,6 +3,8 @@ module KernelAbstractions
 export @kernel
 export @Const, @localmem, @private, @uniform, @synchronize, @index, groupsize
 export Device, GPU, CPU, CUDA, Event
+export async_copy!, pin!
```
lcw:
```suggestion
export async_copy!
```

test/async_copy.jl
```diff
@@ -0,0 +1,25 @@
...
+    event = async_copy!(backend, A, a, dependencies=event)
+    wait(event)
+
+    @test isapprox(a, Array(A))
+    @test isapprox(a, Array(B))
+end
+
+M = 1024
+
+if has_cuda_gpu()
```
leios: Maybe add testsets around these.

Also, I feel there should be a test for all possible copies: CPU-> CPU, GPU->CPU, GPU->GPU (with pin), CPU->GPU.

vchuravy: Interestingly the tests to hit all those code-paths already!

PR #65, opened 2020-03-11
Allow MultiEvent to be created from an Event

lcw:
When building a `MultiEvent` from a `dependencies` argument, it can be
useful to have a `MultiEvent` be created from a single `Event`.


vchuravy:
That's what I get from pushing code quickly. Thanks Lucas.

PR #72, opened 2020-03-20
Implement better error propagation and semaphore

vchuravy:
@lcw observed that this delightful code deadlocks:

```
    barrier = Base.Threads.Event()
    cpu_event = Event(wait, barrier)

    wait(CUDA(), cpu_event) # Event edge on CuDefaultStream
    gpu_event = Event(CUDA()) # Event on CuDefaultStream

    notify(barrier)
    wait(gpu_event)
```

The reason was that I thought it would be okay `wait(CUDA(), cpu_event) # Event edge on CuDefaultStream` to be a blocking event. not thinking about the fact that there could be dependency edges outside our control (here the `barrier`) in Lucas original code `MPI`


vchuravy:
I decided to keep cad3d54 in the history so that we have that code around, I still would prefer that implementation.

cusynchronization.jl
```diff
@@ -0,0 +1,82 @@
...
+    ptr = Base.unsafe_convert(Ptr{T}, sem.sem)
+    unsafe_volatile_store!(ptr, convert(T, val))
+end
+
+@inline function wait(sem::Semaphore{T}) where T
+    while true
+        if sem[] == sem.value
+            break
+        end
+        threadfence_block()
```
vchuravy: @maleadt I am not sure that these `threadfence_block` calls are necessary since no other thread on the device is listening to flag. The code I was inspired by does have them https://github.com/gpudirect/libgdsync/blob/3d0333d6d57af2794088c74f088723f509f794d3/include/gdsync/device.cuh#L112 but they are doing synchronization for another PCI-E device.

PR #74, opened 2020-03-24
Error on return statements inside kernels

### LINKED ISSUE #50: Trailing return yields wrong CPU code.
The `return` statement causes incorrect code.

```
julia> @macroexpand @kernel function f()
         return nothing
       end
```

yields:

```
    function cpu_f(; )
        $(Expr(:aliasscope))
        for var"##I#422" = (KernelAbstractions.__workitems_iterspace)()
            #= /home/vchuravy/src/KernelAbstractions/src/macros.jl:211 =#
            (KernelAbstractions.__validindex)(var"##I#422") || continue
            #= /home/vchuravy/src/KernelAbstractions/src/macros.jl:212 =#
            #= /home/vchuravy/src/KernelAbstractions/src/macros.jl:213 =#
            return nothing
        end
        $(Expr(:popaliasscope))
        return nothing
    end
```

Which means that we will not execute more than the first iteration. Probably we should forbid `return` entirely.

mwarusz:
Fixes #50

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1,10 +1,21 @@
 import MacroTools: splitdef, combinedef, isexpr, postwalk
 
+function find_return(stmt)
+    result = false
+    postwalk(stmt) do expr
+        result |= @capture(expr, return x_)
+        expr
+    end
+    result
+end
+
 # XXX: Proper errors
 function __kernel(expr)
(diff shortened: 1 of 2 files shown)


vchuravy:
Thanks!

PR #78, opened 2020-03-26
Forbid waiting in CUDA on a CPUEvents

lcw:
The current implementation can lead to deadlocks without careful
considerations.  Until this CPU->CUDA synchronization can be done more
robustly, we disable this type of waiting by default.


lcw:
I didn't include the <https://github.com/JuliaGPU/KernelAbstractions.jl/tree/vc/events3> branch in this commit but I can if desired.

PR #80, opened 2020-03-27
Improve launch performance of kernels

vchuravy:
Before:

```
[ Info: Ka Launch
BenchmarkTools.Trial: 
  memory estimate:  1.73 KiB
  allocs estimate:  32
  --------------
  minimum time:     9.702 μs (0.00% GC)
  median time:      10.228 μs (0.00% GC)
  mean time:        11.063 μs (0.00% GC)
  maximum time:     68.124 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

```

After:
```
[ Info: Ka Launch
BenchmarkTools.Trial: 
  memory estimate:  816 bytes
  allocs estimate:  9
  --------------
  minimum time:     4.653 μs (0.00% GC)
  median time:      5.074 μs (0.00% GC)
  mean time:        5.444 μs (0.57% GC)
  maximum time:     317.282 μs (97.24% GC)
  --------------
  samples:          10000
  evals/sample:     7

```

PR #83, opened 2020-03-28
Make CPU kernels synchronous if `dependencies=nothing`

vchuravy:
In order to have them be asynchronous the user can pass in a
`NoneEvent()`. This makes small kernels faster. 

```
~/builds/julia-1.3.1/bin/julia --color=yes --project=. -L ~/perfka.jl
[ Info: Ka + Wait Static
BenchmarkTools.Trial: 
  memory estimate:  80 bytes
  allocs estimate:  3
  --------------
  minimum time:     195.714 ns (0.00% GC)
  median time:      213.559 ns (0.00% GC)
  mean time:        239.629 ns (1.00% GC)
  maximum time:     3.369 μs (86.47% GC)
  --------------
  samples:          10000
  evals/sample:     555

[ Info: Ka Launch
BenchmarkTools.Trial: 
  memory estimate:  816 bytes
  allocs estimate:  9
  --------------
  minimum time:     132.569 ns (0.00% GC)
  median time:      158.833 ns (0.00% GC)
  mean time:        226.427 ns (16.90% GC)
  maximum time:     8.768 μs (96.22% GC)
  --------------
  s amples:          6944
  evals/sample:     893
```

`KA Launch` was measured with `dependencies=NoneEvent()`, script from #80

### DIFF
--- a/src/backends/cpu.jl
+++ b/src/backends/cpu.jl
@@ -95,8 +95,16 @@ function (obj::Kernel{CPU})(args...; ndrange=nothing, workgroupsize=nothing, dep
         ndrange = nothing
     end
 
-    Event(__run, obj, ndrange, iterspace, args, Val(dynamic),
-          dependencies=dependencies, progress=progress)
+    if dependencies === nothing
+        # we can execute this kernel directly, removing the overhead of spawning a task
+        # if you need the the kernel to be executed asynchronously, pass a `NoneEvent` as
+        # a dependency.
+        __run(obj, ndrange, iterspace, args, Val(dynamic))
+        return NoneEvent()
+    else
+        return Event(__run, obj, ndrange, iterspace, args, Val(dynamic),
+              dependencies=dependencies, progress=progress)
+    end
 end
 
 # Inference barriers
--- a/src/backends/cpu.jl
+++ b/src/backends/cpu.jl
@@ -14,7 +14,7 @@ end
 Run function `f` with `args` in a Julia task. If `sticky` is `true` the task
 is run on the thread that launched it.
 """
-function Event(f, args...; dependencies=nothing, progress=yield, sticky=true)
+function Event(f, args...; dependencies=nothing, progress=nothing, sticky=true)
     T = Task() do
         wait(MultiEvent(dependencies), progress)
         f(args...)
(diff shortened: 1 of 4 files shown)


mwarusz:
`ODESolvers` kernels have dependencies on `Event(device)` since they have to synchronize with the default stream on the GPU. Of course we could change dependencies to be `nothing` on the CPU. However, I have to say I am not a fan of using `NoneEvent` to force async on the CPU.  TBH I would prefer having a keyword argument `sync` that would also put kernels on the default stream on the GPU and `assert` that `dependencies === nothing`.

vchuravy:
I would prefer not having synchronous semantics at all, the decision to execute a kernel synchronously would be purely a optimization.
I am not a fan of introducing a keyword, since it will make synchronous semantics mandatory, asserting that dependencies is nothing might help, but I expect an issue quickly where people want to mix synchronous and asynchronous semantics. I do not want the semantics of kernel launches to be even more complicated... 

The reason I choose `NoneEvent` to be the opt-out is so that if people write code like:

```
event = kernel(A)
event = kernel(B. dependencies=event)
wait(event)
```

They at worst only have one synchronous kernel. 

> Of course we could change those dependencies to be nothing on the CPU. 

Interesting. Right now `Event(::CPU)` yields a `NoneEvent()`, we could of course invert the behavior here and make `NoneEvent` or a new `SynchronousEvent` mean that the execution should be happening synchronous. That might be arguably closer to the GPU semantics in the first place.

mwarusz:
Ok. I am just a bit worried that you would be introducing synchronous semantics on the CPU anyway, just using an obscure mechanism to get it. Since running kernels on the CPU that don't have dependencies is probably what a new user will start with, that's the first semantics that they will be exposed to, and they might rely on it. But I guess all this can be made clear in the docs.

vchuravy:
> Ok. I am just a bit worried that you would be introducing synchronous semantics on the CPU anyway, just using an obscure mechanism to get it. Since running kernels on the CPU that don't have dependencies is probably what a new user will start with, that's the first semantics that they will be exposed to, and they might rely on it. 

I am very worried about that. I would be much happier if we don't need this.

PR #85, opened 2020-03-31
Use hostcall for wait and stream GC

vchuravy:
While looking at a profile from @lcw I noticed that
we spend a lot of CPU cycles on `cuEventQuery`. 

The second change is more questionable since it might 
make GPU launches slower (have yet to measure costs).

### DIFF
--- a/lib/CUDAKernels/src/CUDAKernels.jl
+++ b/lib/CUDAKernels/src/CUDAKernels.jl
@@ -75,14 +75,16 @@ import Base: wait
 
 wait(ev::CudaEvent, progress=yield) = wait(CPU(), ev, progress)
 
-function wait(::CPU, ev::CudaEvent, progress=yield)
-    if progress === nothing
-        CUDA.synchronize(ev.event)
-    else
-        while !isdone(ev)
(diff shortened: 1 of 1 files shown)


lcw:
I just tried these commits (on to of master) in CLIMA.  For a 4-GPU run the first commit doesn't change the runtime (wall-clock time per time-step) but the second commit makes the code slightly slower.

vchuravy:
bors try

lib/CUDAKernels/src/CUDAKernels.jl
```diff
@@ -75,14 +75,16 @@ import Base: wait
...
-        while !isdone(ev)
-            progress()
-        end
+function wait(::CPU, ev::CudaEvent, progress=nothing)
+    isdone(ev) && return nothing
+
+    event = Base.Threads.Event()
+    stream = next_stream()
+    wait(CUDADevice(), ev, nothing, stream)
+    CUDA.launch(stream) do
```
vchuravy, line 84:
```suggestion
    CUDA.launch(;stream) do
```

PR #87, opened 2020-04-07
allow docstrings on kernels

simonbyrne:
I wasn't sure what function should have the docstring attached to it.


vchuravy:
Yeah me neither :) That one seems as good as any.

PR #98, opened 2020-05-25
Rework private memory on the CPU

mwarusz:
This PR changes `@private` memory on the CPU to return views to the underlying array with additional dimensions. The main rationale is that it makes passing private arrays to functions more straightforward. For example, currently `@private` vectors in our kernels are passed
as `@views f!(v[:], ...)` which is a bit confusing to the uninitiated. Moreover, this change made is easy to enable scalar `@private` variables.

CC: @jkozdon

### DIFF
--- a/src/backends/cpu.jl
+++ b/src/backends/cpu.jl
@@ -245,3 +245,9 @@ end
 Base.@propagate_inbounds function Base.getindex(A::ScratchArray{N}, I...) where {N}
     return view(A.data, ntuple(_ -> Colon(), Val(N))..., I...)
 end
+Base.@propagate_inbounds function Base.getindex(A::ScratchArray{0}, I...)
+    return A.data[I...]
+end
+Base.@propagate_inbounds function Base.setindex!(A::ScratchArray{0}, val, I...)
+    A.data[I...] = val
+end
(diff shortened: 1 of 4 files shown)


jkozdon:
Awesome. I think this is much cleaner and easier for the uninitiated!

src/backends/cpu.jl
```diff
@@ -242,10 +242,12 @@ Base.eltype(a::ScratchArray) = eltype(a.data)
     return ScratchArray{length(Dims)}(MArray{__size((Dims..., __groupsize(ctx.metadata)...)), T}(undef))
 end
 
-Base.@propagate_inbounds function Base.getindex(A::ScratchArray, I...)
+Base.@propagate_inbounds function Base.getindex(A::ScratchArray{N}, I...) where {N}
+    return @view A.data[ntuple(_ -> Colon(), Val(N))..., I...]
```
vchuravy, line 246: I find `@view` hard to read can you use `view` directly?

mwarusz: Done.

mwarusz:
I realized today that `@views f!(v[:], ...` doesn't even do what I want on the CPU, i.e. it doesn't create a view of the thread-local data. That's because `ScratchArray` is not `<: AbstractArray` and`maybeview` just defaults to `getindex`. Right now, I don't know how to pass a private array to a function and modify it inside.

vchuravy:
Close in favour of #175

PR #103, opened 2020-06-11
fix CuStream deprecation

vchuravy:
bors r+


vchuravy:
As far as I can tell the deprecation was added in CUDAdrv 6.3, which we didn't include in compat bounds, so I am a but bewildered whgy people run into this when using KA from ClimateMachine.

jkozdon:
> so I am a but bewildered whgy people run into this when using KA from ClimateMachine

Because we like to be difficult!

PR #104, opened 2020-06-17
Update KernelAbstractions to use CUDA 1.0

jakebolewski:
Renamed CUDA <: GPU struct to CUDAGPU to avoid name conflicts with CUDA module.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -55,13 +55,13 @@ all(A .== 2.0)
 1. Functions inside kernels are forcefully inlined, except when marked with `@noinline`.
 2. Floating-point multiplication, addition, subtraction are marked contractable.
 
-## Important differences to CUDA
+## Important differences to CUDA.jl
 
 1. The kernels are automatically bounds-checked against either the dynamic or statically
    provided `ndrange`.
 2. Functions like `Base.sin` are mapped to `CUDA.sin`.
 
-## Important differences to GPUifyLoops
+## Important differences to GPUifyLoops.jl
 
 1. `@scratch` has been renamed to `@private`, and the semantics have changed. Instead
    of denoting how many dimensions are implicit on the GPU, you only ever provide the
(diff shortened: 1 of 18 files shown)


jakebolewski:
Part of https://github.com/CliMA/ClimateMachine.jl/issues/1228

docs/src/index.md
```diff
@@ -55,11 +55,11 @@ all(A .== 2.0)
 1. Functions inside kernels are forcefully inlined, except when marked with `@noinline`.
 2. Floating-point multiplication, addition, subtraction are marked contractable.
 
-## Important differences to CUDAnative
+## Important differences to CUDA
```
vchuravy, line 58:
```suggestion
## Important differences to CUDA.jl
```

src/KernelAbstractions.jl
```diff
@@ -303,7 +303,7 @@ abstract type Device end
 abstract type GPU <: Device end
 
 struct CPU <: Device end
-struct CUDA <: GPU end
+struct CUDAGPU <: GPU end
```
vchuravy, line 306: I am not a fan of the name, maybe we rename it to `NVIDIA` like the prospective names below?

But I don't have a super strong opinion on this.

jakebolewski: yeah I have no issues changing it, I'll switch to the consensus.

NVIDIA wouldn't clash with CUDA and aligns with the other names so it seems like a good candidate.

jakebolewski: I was going to use `CUDABackend` but I'm not sure that is better as it's a little long

lcw: What names are you thinking about for the AMD and Intel GPU backends?

jakebolewski: it seems `AMD` and `Intel`?

jakebolewski: There are a lot of ` device isa CUDA` in the CLIMA code, so would `CUDADevice` be a better name?

lcw: Either way sounds okay with me.

jakebolewski:
Renamed to `CUDADevice`.

jakebolewski:
@vchuravy can you tag and register another version?

PR #106, opened 2020-07-02
Move CUDA backend to CUDA.jl

vchuravy:
In order to use KernelAbstractions to implement GPUArrays the
GPU backends need to be implemented in CUDA.jl and ROCM.jl, otherwise
we could get circular dependencies.

TODO:
- Make tests work properly


vchuravy:
Closed in favour of #200

PR #130, opened 2020-09-11
Rewrite expoenent implementation for the CUDA backend with Cassette for Julia 1.5 compatibility

jakebolewski:


vchuravy:
Thanks for figuring that out!

PR #131, opened 2020-09-16
@ka_code_typed implementation for KernelAbstractions

mjulian31:
@vchuravy Here is my implementation of @ka_code_typed.


Manifest.toml
```diff
@@ -0,0 +1,199 @@
+# This file is machine-generated - editing it directly is not advised
```
vchuravy, line 1: I don't check in Manifest.toml's normally, since they are version dependent and can contain paths to local files.

src/backends/cpu.jl
```diff
@@ -78,31 +78,38 @@ function async_copy!(::CPU, A, B; dependencies=nothing, progress=nothing)
...
+
+    if length(blocks(iterspace)) == 0
+        return MultiEvent(dependencies)
+    end
+
+    Event(__run, obj, ndrange, iterspace, args, dynamic,
+          dependencies=dependencies, progress=progress)
+end
+
+function launch_config(kernel::Kernel{CPU}, ndrange=nothing, workgroupsize=nothing)
```
vchuravy, line 95:
```suggestion
function launch_config(kernel::Kernel{CPU}, ndrange, workgroupsize)
```

src/reflection.jl
```diff
@@ -0,0 +1,34 @@
+import InteractiveUtils
+export @ka_code_typed
+
+function ka_code_typed(kernel, argtypes; ndrange=nothing, workgroupsize=nothing)
```
vchuravy, line 4:
```suggestion
function ka_code_typed(kernel, argtypes; ndrange=nothing, workgroupsize=nothing, dependencies=nothing, kwargs...)
```

src/reflection.jl
```diff
@@ -0,0 +1,34 @@
...
+    # get the first block
+    block = @inbounds KernelAbstractions.blocks(iterspace)[1]
+    # get a context of the kernel based on the first block
+    ctx = KernelAbstractions.mkcontext(kernel, block, ndrange, iterspace, dynamic)
+    # reformat
+    if argtypes isa Type
+        argtypes = argtypes.parameters
+    end
+    # use code_typed
+    return InteractiveUtils.code_typed(KernelAbstractions.Cassette.overdub, (typeof(ctx), typeof(kernel.f), argtypes...))
```
vchuravy, line 16:
```suggestion
    return InteractiveUtils.code_typed(KernelAbstractions.Cassette.overdub, (typeof(ctx), typeof(kernel.f), argtypes...); kwargs...)
```

`code_typed` takes a couple of key-word arguments, that control the output.

src/reflection.jl
```diff
@@ -0,0 +1,34 @@
...
+end
+
+
+macro ka_code_typed(ex0...)
```
vchuravy, line 20: Can you add a docstring as well?

Ideally `@ka_code_typed` would take the same arguments as `code_typed`. For `code_typed` that is just `optimize=false`. As an example `@code_typed optimize=false sin(1.0)`.

src/reflection.jl
```diff
@@ -0,0 +1,34 @@
...
+    if length(ex0) == 1
+        kw = ex0[1].args[end]
```
vchuravy, line 22: This only works for one argument? What about providing both `workgroupsize` and `ndrange`?

vchuravy (commented):
Excellent! This is looking good. Can you add a test?

Ideally we also would have an interactive mode like, https://github.com/JuliaGPU/GPUCompiler.jl/blob/177bff39877f39c925e340cef391c1e11bfa28c4/src/reflection.jl#L48-L56

src/reflection.jl
mjulian31: If there is only one argument, it is because they called it with @ka_code_typed kernel(CPU(), ndrange=8) for example. They can also call it with @ka_code_typed ndrange=8 kernel(CPU()), but that might not be necessary to support.

Manifest.toml
mjulian31: Ah ok, I wasn't sure if it was needed or not.

src/reflection.jl
mjulian31: > Excellent! This is looking good. Can you add a test?
> 
> Ideally we also would have an interactive mode like, https://github.com/JuliaGPU/GPUCompiler.jl/blob/177bff39877f39c925e340cef391c1e11bfa28c4/src/reflection.jl#L48-L56

I can look into adding the interactive mode. I wasn't sure if that was the goal for this or if was only to have a functioning method.
I can also add a test, but I'm not sure what I should check. Just that it functions for a few different kind of calls, or that it produces a specific output? If you could link me to the @code_typed tests, I could try to replicate those to some degree.

src/reflection.jl
mjulian31: Yes I can add a doc string and look into adding the optimize argument. Is there an example of where optimize gets used when passing into code_typed()?

src/reflection.jl
vchuravy: Yeah something simple like https://github.com/JuliaLang/julia/blob/b884dcad9598f862ab6c08c01e2800942769470c/stdlib/InteractiveUtils/test/runtests.jl#L127

But where you test the different KA keyword arguments. 

> I can look into adding the interactive mode. I wasn't sure if that was the goal for this or if was only to have a functioning method.

Interactive mode is definitely a stretch goal.

src/reflection.jl
vchuravy: `code_typed(sin, Tuple{Float64}, optimize=false)`

src/reflection.jl
vchuravy: Right, but what about `@ka_code_typed kernel(args...; ndrange=128, workgroupsize=8)`?

mjulian31: In that instance, I believe the whole kernel is the one argument ex0 that has different parameters inside of it. However, I think that I may need to split up in a similar way to get workgroup size to work for example

PR #150, opened 2020-11-18
Add locking to stream GC

jpsamaroo:


src/backends/cuda.jl
```diff
@@ -20,26 +20,28 @@ end
 # Possible improvements
 # - Add a background task that occasionally scans all streams
 # - Add a hysterisis by checking a "since last scanned" timestamp
-# - Add locking
+const STREAM_GC_LOCK = Threads.ReentrantLock()
 function next_stream()
-    if !isempty(FREE_STREAMS)
-        return pop!(FREE_STREAMS)
-    end
+    lock(STREAM_GC_LOCK) do
```
vchuravy, line 25: `lock` does a `try ... finally` right?

vchuravy (approved):

src/backends/cuda.jl
jpsamaroo: Correct

PR #175, opened 2021-01-07
Fancy private

vchuravy:
Support for:

```
@kernel function new_private()
    @private mem = 1
    mem
end
```

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -117,6 +117,7 @@ struct WorkgroupLoop
     indicies :: Vector{Any}
     stmts :: Vector{Any}
     allocations :: Vector{Any}
+    private_allocations :: Vector{Any}
     private :: Vector{Any}
 end
 
@@ -146,13 +147,15 @@ function split(stmts,
 
     current     = Any[]
     allocations = Any[]
+    private_allocations = Any[]
     new_stmts   = Any[]
     for stmt in stmts
         has_sync = find_sync(stmt)
         if has_sync
-            loop = WorkgroupLoop(deepcopy(indicies), current, allocations, deepcopy(private))
+            loop = WorkgroupLoop(deepcopy(indicies), current, allocations, private_allocations, deepcopy(private))
             push!(new_stmts, emit(loop))
             allocations = Any[]
+            private_allocations = Any[]
             current     = Any[]
             is_sync(stmt) && continue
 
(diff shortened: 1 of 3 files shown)


vchuravy:
@mwarusz this might also fix you use-case from #98

vchuravy:
we should turn `mem = @private T dims` into `@private mem = ScratchArray{T, dims}`

vchuravy:
bors try

src/macros.jl
```diff
@@ -212,13 +226,34 @@ function emit(loop)
...
     stmts = Any[]
     append!(stmts, loop.allocations)
+
+    # private_allocations turn into lhs = ntuple(i->rhs, length(__workitems_iterspace()))
+    N = gensym(:N)
+    push!(stmts, :($N = length($__workitems_iterspace())))
+
+    for stmt in loop.private_allocations
+        if @capture(stmt, lhs_ = rhs_)
+            push!(stmts, :($lhs = ntuple(_->$rhs, $N)))
```
vchuravy, line 236: @simonbyrne you asked if we could support something like this:

```
e = @index(Group, Linear)
i, j = @index(Local, NTuple)
@private mem = Fprog[i,j,:,e]
```

We would need to scan `rhs` for any uses of `loop.indicies` and then project `i`/`_` onto `__workitems_iterspace` inside to obtain a valid `idx` and then instantiate the index expressions `$(loop.indicies...)`. Doable, but pfff.

simonbyrne: In second thought, I don't think this is really necessary.

simonbyrne:
it would be useful to support loop indices in `@private`, i.e.
```
i, j = @index(Local, NTuple)
@private f = foo(i)
```

simonbyrne:
bors try

simonbyrne:
> it would be useful to support loop indices in `@private`, i.e.
> 
> ```
> i, j = @index(Local, NTuple)
> @private f = foo(i)
> ```

@vchuravy any chance we could support this?

vchuravy:
I thought you said it isn't necessary https://github.com/JuliaGPU/KernelAbstractions.jl/pull/175/files#r553653540 ;)

simonbyrne:
oh, i thought that was about something else.

Why you do the same thing you do for loops?

PR #206, opened 2021-02-07
tune CUDA kernels automatically

simeonschaub:
This is still quite rough around the edges, but I am putting this up for feedback. This automatically splits up the threads over leading dimensions of the ndrange for better performance if the first dimension is small.

### DIFF
--- a/src/backends/cuda.jl
+++ b/src/backends/cuda.jl
@@ -144,51 +144,43 @@ function launch_config(kernel::Kernel{CUDADevice}, ndrange, workgroupsize)
         workgroupsize = (workgroupsize, )
     end
 
-    if KernelAbstractions.workgroupsize(kernel) <: DynamicSize && workgroupsize === nothing
-        # TODO: allow for NDRange{1, DynamicSize, DynamicSize}(nothing, nothing)
-        #       and actually use CUDA autotuning
-        workgroupsize = (256,)
-    end
-
     # partition checked that the ndrange's agreed
     if KernelAbstractions.ndrange(kernel) <: StaticSize
         ndrange = nothing
     end
 
-    iterspace, dynamic = partition(kernel, ndrange, workgroupsize)
+    iterspace, dynamic = if KernelAbstractions.workgroupsize(kernel) <: DynamicSize &&
+        workgroupsize === nothing
+        # use ndrange as preliminary workgroupsize for autotuning
+        partition(kernel, ndrange, ndrange)
+    else
+        partition(kernel, ndrange, workgroupsize)
+    end
 
     return ndrange, workgroupsize, iterspace, dynamic
 end
 
 function threads_to_workgroupsize(threads, ndrange)
-    workgroupsize = ones(Int, length(ndrange))
-    workgroupsize[1] = threads
-    i = 1
-    for outer i in 1:length(ndrange)-1
-        workgroupsize[i] > ndrange[i] || break
-        workgroupsize[i+1] = div(threads, prod(ndrange[1:i]))
-        workgroupsize[i] = ndrange[i]
+    total = 1
+    return map(ndrange) do n
+        x = min(div(threads, total), n)
+        total *= x
+        return x
     end
-    workgroupsize[i] = min(workgroupsize[i], ndrange[i])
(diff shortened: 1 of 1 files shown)


simeonschaub:
Trying this out on my real example, it seems like this adds quite a lot of overhead, so the optimal thread size should probably be determined at kernel creation time, not every time it is invoked.

simeonschaub:
Turns out that this overhead came from having to switch from a static to a dynamic workgroupsize, so I am thinking that it might be better to have a separate API for determining the optimal workgroupsize for a kernel.

simeonschaub:
Well, that was embarassing... Turns out the weird performance issues I was seeing came from me accidentally capping the CUDA memory limit way to low. :facepalm: With that fixed, this does actually give a nice speedup even if I change to a dynamic workgroupsize.

src/backends/cuda.jl
```diff
@@ -160,9 +160,36 @@ function launch_config(kernel::Kernel{CUDADevice}, ndrange, workgroupsize)
...
+end
+
 function (obj::Kernel{CUDADevice})(args...; ndrange=nothing, dependencies=nothing, workgroupsize=nothing, progress=yield)
 
-    ndrange, workgroupsize, iterspace, dynamic = launch_config(obj, ndrange, workgroupsize)
+    ndrange, _workgroupsize, iterspace, dynamic = launch_config(obj, ndrange, workgroupsize)
+    kernel = nothing
+
+    if KernelAbstractions.workgroupsize(obj) <: DynamicSize && workgroupsize === nothing
+        ndrange, _, iterspace, _ = launch_config(obj, ndrange, ndrange)
```
vchuravy, line 182: Why can't we re-use the result of `launch_config` three lines above?

src/backends/cuda.jl
```diff
@@ -160,9 +160,36 @@ function launch_config(kernel::Kernel{CUDADevice}, ndrange, workgroupsize)
...
+        ctx = mkcontext(obj, ndrange, iterspace)
+        kernel = CUDA.@cuda name=String(nameof(obj.f)) launch=false Cassette.overdub(ctx, obj.f, args...)
```
vchuravy, line 184:
```suggestion
        kernel = CUDA.@cuda launch=false name=String(nameof(obj.f))  Cassette.overdub(ctx, obj.f, args...)
```

vchuravy:
bors try

src/backends/cuda.jl
simeonschaub: The problem is that the type of `workgroupsize` is statically embedded in the cassette pass, so to be able to use the already compiled `cufunction`, the preliminary `workgroupsize` we pass here has the same number of dimensions than our eventual optimized `workgroupsize`. Since the actual value doesn't really matter, I just used `ndrange` here, and then `threads_to_workgroupsize` always returns a tuple of the same length as `ndrange`. My current implementation is structured a bit weird, I will try to make this a little more obvious.

simeonschaub:
Ok, I think I am fairly happy with this now. If you agree with the way this now works, it should be good to go from my side.

src/backends/cuda.jl
```diff
@@ -144,25 +144,44 @@ function launch_config(kernel::Kernel{CUDADevice}, ndrange, workgroupsize)
...
+    kernel = nothing
+
+    if KernelAbstractions.workgroupsize(obj) <: DynamicSize && workgroupsize === nothing
+        ctx = mkcontext(obj, ndrange, iterspace)
+        kernel = CUDA.@cuda launch=false name=String(nameof(obj.f))  Cassette.overdub(ctx, obj.f, args...)
+        config = CUDA.launch_configuration(kernel.fun; max_threads=prod(ndrange))
+
+        workgroupsize = threads_to_workgroupsize(config.threads, ndrange)
+        iterspace, dynamic = partition(obj, ndrange, workgroupsize)
+    end
```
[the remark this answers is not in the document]
simeonschaub: Ah, true

simeonschaub:
Mind also tagging a release with this?

vchuravy:
Done https://github.com/JuliaRegistries/General/pull/29752

PR #207, opened 2021-02-11
More robust test/examples.jl

tkf:
This is a PR onto #200.

I think the test is failing because `Base.current_project()` doesn't work nicely with `--project`:

```
arakaki@amdci2 ~/.j/d/KernelAbstractions (vc/split)> pwd
/home/arakaki/.julia/dev/KernelAbstractions
arakaki@amdci2 ~/.j/d/KernelAbstractions (vc/split)> julia --startup-file=no --project=test -e 'display(Base.current_project())'
"/home/arakaki/.julia/dev/KernelAbstractions/Project.toml"⏎
arakaki@amdci2 ~/.j/d/KernelAbstractions (vc/split)> julia --startup-file=no --project=test -e 'display(Base.load_path())'
3-element Array{String,1}:
 "/home/arakaki/.julia/dev/KernelAbstractions/test/Project.toml"
 "/home/arakaki/.julia/environments/v1.5/Project.toml"
 "/home/arakaki/opt/julia/julia-1.5.2/share/julia/stdlib/v1.5"⏎
```

A quick workaround may be to use `Base.load_path_setup_code()` to propagate load path etc to the child processes.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -23,7 +23,7 @@ steps:
     commands:
       - julia --project=test -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()))'
       - julia --project=test -e 'using Pkg; Pkg.develop(PackageSpec(path=joinpath(pwd(),"lib","CUDAKernels")))'
-      - JULIA_LOAD_PATH=@ julia --project=test --color=yes --check-bounds=yes --code-coverage=user --depwarn=yes test/runtests.jl
+      - julia --project=test --color=yes --check-bounds=yes --code-coverage=user --depwarn=yes test/runtests.jl
     agents:
       queue: "juliagpu"
       cuda: "*"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -8,9 +8,7 @@ steps:
     commands:
       - julia --project=test -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()))'
       - julia --project=test -e 'using Pkg; Pkg.develop(PackageSpec(path=joinpath(pwd(),"lib","CUDAKernels")))'
-      - julia -e 'using Pkg; Pkg.add(name="Run", version="0.1")'
-      - julia -e 'using Run; Run.prepare_test()'
-      - julia -e 'using Run; Run.test()'
+      - julia --project=test --color=yes --check-bounds=yes --code-coverage=user --depwarn=yes test/runtests.jl
     agents:
       queue: "juliagpu"
       cuda: "*"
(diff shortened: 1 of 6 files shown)


vchuravy:
bors try

vchuravy:
Fascinatingly now it is only partly broken

tkf:
Maybe the failure in Windows is due to the incompatibility between powershell and bash?

https://github.com/JuliaGPU/KernelAbstractions.jl/runs/1882919388?check_suite_focus=true#step:6:3

I just added the patch that is supposed to use bash everywhere.

I'm not sure why it'd fail on ubuntu, though https://github.com/JuliaGPU/KernelAbstractions.jl/runs/1882919107#step:9:128

I can't repro it locally

vchuravy:
bors try

vchuravy:
bors help

vchuravy:
bors delegate=@tkf

DilumAluthge:
I think it's 

bors delegate=tkf

tkf:
Thanks, Dilum!

tkf:
bors try

tkf:
bors try

tkf:
bors try

tkf:
bors try

tkf:
bors try

DilumAluthge:
Why all this business with the load path? Why not just do `Pkg.test("KernelAbstractions")`?

vchuravy:
Ha! If it were just that simple, `Pkg.test` doesn't like the mono-repo setup we are going for.

tkf:
So, it looks like this is a Julia 1.6-specific problem https://github.com/JuliaLang/julia/issues/39631

(I think this is orthogonal to the issue that Pkg.test is not enough for this mono-repo setting.)

vchuravy:
right you need to add `@stdlib`

tkf:
I believe the behavior of 1.6 is a bug. You shouldn't need anything in the top-level dependencies for transitive dependencies. If it's in the manifest, it should work.

tkf:
OK, github action worked in my repo https://github.com/tkf/KernelAbstractions.jl/pull/1 (esp https://github.com/tkf/KernelAbstractions.jl/runs/1883606478) so I think this is good to go.

I'm not merging via bors since I don't know if it supports merging to a non-default branch.

vchuravy:
Thanks Taka!

DilumAluthge:
> Ha! If it were just that simple, `Pkg.test` doesn't like the mono-repo setup we are going for.

I'm not sure I follow. You should be able to do the following:

```bash
julia test.jl
```

Where the contents of `test.jl` are:
```julia 
import Pkg

# KernelAbstractions
Pkg.develop(Pkg.PackageSpec(path = pwd()))

# CUDAKernels
Pkg.develop(Pkg.PackageSpec(path = joinpath(pwd(), "lib", "CUDAKernals")))

Pkg.test(["KernelAbstractions", "CUDAKernels"])
```

Does that not work? If it doesn't work, we should patch Pkg to make it work.

PR #208, opened 2021-02-11
performant matmul example for KA

### LINKED ISSUE #205: Feature request: performant matmul! example
I've noticed that `transpose` and `copy` are implemented in both a naive and a performant way [here](https://juliagpu.github.io/KernelAbstractions.jl/dev/examples/performance/). On the other hand, for `matmul!`, I could only find the naive implementation [here](https://juliagpu.github.io/KernelAbstractions.jl/dev/examples/matmul/).

I understand that a performant julia `matmul!` on the GPU is a whole research project on its own over at [GemmKernels.jl](https://github.com/JuliaGPU/GemmKernels.jl), but I was curious whether there would be some "middle ground" that loses some performance compared to cuBLAS but is easy and general. IMO, a good endgoal would be to be able to generate that code from a DSL like Tullio (see https://github.com/mcabbott/Tullio.jl/issues/80#issuecomment-770614929).

mjulian31:
Fixes #205

### DIFF
--- a/examples/performant_matmul.jl
+++ b/examples/performant_matmul.jl
@@ -77,7 +77,6 @@ C = zeros(N, M)
 
 kern = coalesced_matmul_kernel!(CPU(), (TILE_DIM, TILE_DIM))
 
-@btime wait(kern(C, $A, $B, $N, $R, $M, ndrange=size(C))) setup=(C = zeros(N, M))
 
 wait(kern(C, A, B, N, R, M, ndrange=size(C)))
 @test isapprox(A*B, C)
--- a/examples/performant_matmul.jl
+++ b/examples/performant_matmul.jl
@@ -1,0 +1,85 @@
+using BenchmarkTools
+using KernelAbstractions
+using StaticArrays
+using Test
+
+const TILE_DIM = 32
+
+@kernel function coalesced_matmul_kernel!(output, @Const(input1), @Const(input2), N, R, M,
+                                             ::Val{BANK}=Val(1)) where BANK
+     gi, gj = @index(Group, NTuple)
+     i, j   = @index(Local, NTuple)
+
+     TILE_DIM = @uniform groupsize()[1]
+
+     # +1 to avoid bank conflicts on shared memory
+     tile1 = @localmem eltype(output) (TILE_DIM+BANK, TILE_DIM)
+     tile2 = @localmem eltype(output) (TILE_DIM+BANK, TILE_DIM)
+
+     # private variable for tile output
+     outval = @private eltype(output) 1
+     @inbounds outval[1] = -zero(eltype(output))
+
+     @uniform N = size(output, 1)
+     # number of tiles depends on inner dimension
+     @uniform NUM_TILES = div(R + TILE_DIM - 1, TILE_DIM)
+
+     # loop over all tiles needed for this calculation
+     for t in 0:NUM_TILES-1
+         # Can't use @index(Global), because we use a smaller ndrange
+         I = (gi-1) * TILE_DIM + i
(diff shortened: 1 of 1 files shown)


vchuravy:
bors try

examples/performant_matmul.jl
```diff
@@ -0,0 +1,84 @@
+using BenchmarkTools
+using KernelAbstractions
+using StaticArrays
```
vchuravy, line 3:
```suggestion
using StaticArrays
using Test
```

vchuravy:
bors try

DilumAluthge:
Will also need to add BenchmarkTools.

examples/performant_matmul.jl
```diff
@@ -0,0 +1,85 @@
+using BenchmarkTools
```
vchuravy, line 1:
```suggestion
```

examples/performant_matmul.jl
```diff
@@ -0,0 +1,85 @@
...
+R = 512
+M = 2048
+A = rand(N, R)
+B = rand(R, M)
+C = zeros(N, M)
+
+kern = coalesced_matmul_kernel!(CPU(), (TILE_DIM, TILE_DIM))
+
+wait(kern(C, A, B, N, R, M, ndrange=size(C)))
+@btime wait(kern(C, $A, $B, $N, $R, $M, ndrange=size(C))) setup=(C = zeros(N, M))
```
vchuravy, line 82:
```suggestion
```

examples/performant_matmul.jl
```diff
@@ -0,0 +1,85 @@
...
+N = 1024
+R = 512
+M = 2048
+A = rand(N, R)
+B = rand(R, M)
+C = zeros(N, M)
+
+kern = coalesced_matmul_kernel!(CPU(), (TILE_DIM, TILE_DIM))
+
+wait(kern(C, A, B, N, R, M, ndrange=size(C)))
```
vchuravy, line 81:
```suggestion
```

vchuravy:
bors try

DilumAluthge:
@vchuravy Is there a particular reason that you took out the benchmarking?

DilumAluthge:
bors try

vchuravy:
> @vchuravy Is there a particular reason that you took out the benchmarking?

Don't want to eat up CI time.

examples/performant_matmul.jl
```diff
@@ -0,0 +1,82 @@
...
+     @inbounds outval[1] = -zero(eltype(output))
+
+     @uniform N = size(output, 1)
+     # number of tiles depends on inner dimension
+     @uniform NUM_TILES = div(R + TILE_DIM - 1, TILE_DIM)
+
+     # loop over all tiles needed for this calculation
+     for t in 0:NUM_TILES-1
+         # Can't use @index(Global), because we use a smaller ndrange
+         I = (gi-1) * TILE_DIM + i
```
piever, line 29: I'm sure there is a reason for this, but I'm not sure I understand it. Is there an obvious reason why `I = (gi-1) * TILE_DIM + i` here could actually give a different result than `I = (gi-1) * TILE_DIM + i` a few lines below? Is the idea that one should always recompute `I, J` before using them because `i, gi, j, gj` may have changed?

vchuravy:
bors try

examples/performant_matmul.jl
vchuravy: It needs to be recalculated after the `@synchronize` otherwise you run into scope issues. This is the kind of problem KernelCompiler.jl will hopefully be able to solve.

PR #211, opened 2021-02-15
Test suite: add a newline at the end of a print statement

DilumAluthge:
This makes the test suite output a little easier to read.

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -29,4 +29,4 @@ status = [
   "Documentation",
 ]
 
-timeout_sec = 3600
+timeout_sec = 10800
--- a/test/print_test.jl
+++ b/test/print_test.jl
@@ -24,6 +24,6 @@ end
     wait(test_print(CPU()))
     @test true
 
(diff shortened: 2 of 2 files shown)


DilumAluthge:
bors merge

bors.toml
```diff
@@ -29,4 +29,4 @@ status = [
   "Documentation",
 ]
 
-timeout_sec = 3600
```
vchuravy, line 32: Why did you bump the timeout?

DilumAluthge: If all of the Buildkite runners are in use, Bors might need to wait a while. This doesn't count towards the Buildkite timeout - once the Buildkite job starts, it has a timeout of 60 minutes. But the time waiting for a Buildkite runner to become available does count towards the Bors timeout. So if all the Buildkite runners are in use, we might hit the Bors timeout of 60 minutes. So I increased it to 180 minutes.

We can decrease it again if you want.

PR #219, opened 2021-03-05
Combine tests into parameterized testsuite

jpsamaroo:


vchuravy:
bors try

test/test.jl
```diff
@@ -68,82 +64,73 @@ end
...
-            AT = CUDA.CuDeviceArray{Float32, 2, CUDA.AS.Global}
-            IR = sprint() do io
-                CUDA.code_llvm(io, KernelAbstractions.Cassette.overdub,
-                               (typeof(ctx), typeof(kernel.f), AT, AT),
-                               kernel=true, optimize=true)
-            end
+        if backend == CPU
+            @test occursin("!alias.scope", IR)
+            @test occursin("!noalias", IR)
+        else
```
vchuravy, line 132:
```suggestion
        elseif backend == CUDA
```

jpsamaroo: Yeah I have something similar locally. We can't directly compare the backend (since `CUDADevice` might not exist when `ROCKernels is tested in the future), so I currently just check the relevant env. var.

jpsamaroo:
CPU and CUDA pass for me locally. Let's see if this works:

bors try

jpsamaroo:
Not really sure what to do about the fact that a CUDA runner doesn't have a CUDA GPU...

DilumAluthge:
@maleadt

DilumAluthge:
bors try

DilumAluthge:
Hmmm. It seems to be happening on a couple different runners. Is there any chance there's a bug in `has_cuda_gpu()`?

DilumAluthge:
Maybe, before calling `has_cuda_gpu()`, we could print the output of `CUDA.versioninfo()`.

maleadt:
> Not really sure what to do about the fact that a CUDA runner doesn't have a CUDA GPU...

Those runners do have a GPU. On CI, you should probably call `functional` with `verbose=true`? Anyway, this is probably a bug in CUDA.jl, and I think it's because the CUDA 11.2 installation on those runners is not being detected by the package. CUDA.jl#master does support that version, but it also should just fall back to an artifact installation if the local CUDA set-up is unrecognized, and that's what https://github.com/JuliaGPU/CUDA.jl/pull/740 does. For now, you can force that by setting `JULIA_CUDA_USE_BINARYBUILDER=false`.

.buildkite/pipeline.yml
```diff
@@ -14,6 +14,8 @@ steps:
     agents:
       queue: "juliagpu"
       cuda: "*"
+    env:
```
vchuravy, line 17:
```suggestion
    env:
      JULIA_CUDA_USE_BINARYBUILDER=false
```

.buildkite/pipeline.yml
```diff
@@ -14,6 +14,9 @@ steps:
...
+      JULIA_CUDA_USE_BINARYBUILDER=false
```
vchuravy, line 18:
```suggestion
      JULIA_CUDA_USE_BINARYBUILDER: "false"
```

.buildkite/pipeline.yml
```diff
@@ -31,6 +34,9 @@ steps:
...
```
vchuravy, line 38:
```suggestion
      JULIA_CUDA_USE_BINARYBUILDER: "false"
```

.buildkite/pipeline.yml
```diff
@@ -48,6 +54,9 @@ steps:
...
```
vchuravy, line 58:
```suggestion
      JULIA_CUDA_USE_BINARYBUILDER: "false"
```

vchuravy:
bors try

jpsamaroo:
Don't we also need CUDA#master?

maleadt:
No, the env var should disable preferred use of local CUDA on CI: https://github.com/JuliaGPU/CUDA.jl/blob/634f841879bc02dbc618286808ae09427b10b8bc/deps/bindeps.jl#L374

vchuravy:
bors try

PR #220, opened 2021-03-10
Adds cosd and sind to list of cudafuns

glwagner:
cosd and sind are like cos and sin, except their arguments have units of "degrees" rather than radians.

Is there anything else I should do? Bump ~~minor~~ patch version for tagging purposes?

### DIFF
--- a/lib/CUDAKernels/src/CUDAKernels.jl
+++ b/lib/CUDAKernels/src/CUDAKernels.jl
@@ -285,16 +285,16 @@ KernelAbstractions.generate_overdubs(@__MODULE__, CUDACtx)
 @inline Cassette.overdub(::CUDACtx, ::typeof(^), x::Union{Float32, Float64}, y::Int64) = CUDA.pow(x, y)
 
 # libdevice.jl
-const cudafuns = (:cos, :cospi, :sin, :sinpi, :tan,
-          :acos, :asin, :atan,
-          :cosh, :sinh, :tanh,
-          :acosh, :asinh, :atanh,
-          :log, :log10, :log1p, :log2,
-          :exp, :exp2, :exp10, :expm1, :ldexp,
-          # :isfinite, :isinf, :isnan, :signbit,
-          :abs,
-          :sqrt, :cbrt,
-          :ceil, :floor,)
+const cudafuns = (:cos, :cospi, :cosd, :sin, :sinpi, :sind, :tan,
+                  :acos, :asin, :atan,
+                  :cosh, :sinh, :tanh,
+                  :acosh, :asinh, :atanh,
+                  :log, :log10, :log1p, :log2,
+                  :exp, :exp2, :exp10, :expm1, :ldexp,
+                  # :isfinite, :isinf, :isnan, :signbit,
+                  :abs,
+                  :sqrt, :cbrt,
+                  :ceil, :floor,)
 for f in cudafuns
     @eval function Cassette.overdub(ctx::CUDACtx, ::typeof(Base.$f), x::Union{Float32, Float64})
         @Base._inline_meta


vchuravy:
Can you add a test? And no need to bump the version since we haven't tagged 0.6 yet.

ali-ramadhan:
Version bump (v0.5.4?) might be nice so Oceananigans.jl can depend on a tagged release.

vchuravy:
> Version bump (v0.5.4?) might be nice so Oceananigans.jl can depend on a tagged release.

Then we need to backport this, e.g. branch of https://github.com/JuliaGPU/KernelAbstractions.jl/commit/a6e49655b600174a41c9aad798a6f72e0e955321 and target the commit there.

glwagner:
Because bumping again would produce too many versions?

vchuravy:
no because master is the development branch for 0.6.0 so right now there are changes queued up that are backwards incompatible.

glwagner:
Ah, makes sense.

glwagner:
Since we have a temporary solution (we can just manually convert the arguments of `sin` and `cos` to radians we probably don't need this until 0.6.0 so maybe sticking with this PR is ok.

glwagner:
As for a test, an integration-type test would launch a kernel that uses each `cudafun` for `Float32` and `Float64`. Is that in scope? Or is there are simpler way to test that `sind` compiles to `CUDA.sind`?

vchuravy:
Integration test is fine, just a simple example kernel.

vchuravy:
@glwagner can you rebase your commit on the branch `release-0.5`?

vchuravy:
Are you sure that CUDA.jl defines `cosd` and `sind`?

glwagner:
There's this:

```julia
julia> using CUDA

help?> CUDA.sind
?  sind(x)

  Compute sine of x, where x is in degrees. 

help?> CUDA.cosd
  cosd(x)

  Compute cosine of x, where x is in degrees. 
```

I can rebase --- I'm on vacation now, so give me a few days!

vchuravy:
That is I think just the Base implementation...

glwagner:
@vchuravy does this show that the implementations are special? Not sure how to check...

```julia
julia> using CUDA, BenchmarkTools

julia> calc_sind(x) = CUDA.@sync CUDA.sind.(x)
calc_sind (generic function with 1 method)

julia> calc_sin(x) = CUDA.@sync CUDA.sin.(x)
calc_sin (generic function with 1 method)

julia> x = CuArray(rand(4096, 4096));

julia> @btime calc_sin(x);
  505.311 μs (34 allocations: 848 bytes)

julia> @btime calc_sind(x);
  502.704 μs (35 allocations: 880 bytes)

**Comment by vchuravy** [comment]
```
julia> methods(CUDA.sind)
# 2 methods for generic function "sind":
[1] sind(x::Real) in Base.Math at special/trig.jl:1182
[2] sind(z) in Base.Math at special/trig.jl:1268

julia> methods(CUDA.sin)
# 2 methods for generic function "sin":
[1] sin(x::Float64) in CUDA at /home/vchuravy/.julia/packages/CUDA/qEV3Y/src/device/intrinsics/math.jl:12
[2] sin(x::Float32) in CUDA at /home/vchuravy/.julia/packages/CUDA/qEV3Y/src/device/intrinsics/math.jl:13
```

So `CUDA.sind` is just `Base.sind`. This just shows that Julia implements `sind` manually, and that the implementation seems to work on the GPU. I am positively surprised by the speed though. 

cc: @maleadt

jpsamaroo:
It would be great if you could also add these definitions to ROCKernels!

maleadt:
Latest GPUCompiler can do contextual dispatch, and a compatible CUDA release is for this or next week, so most of this code will be able to go then.

glwagner:
Closing since if I understand @maleadt correctly the future is bright and we've no use for `cudafuns` 🚀

PR #222, opened 2021-03-17
[CUDAKernels] add an implicit sync to kernels with no dependencies

### LINKED ISSUE #221: Problem with a simple add vecs for some arrays sizes on CUDA
So I am not sure what I am doing wrong:

```julia
using KernelAbstractions
using CUDA
using Printf
using Random

@kernel function ka_knl_add!(C, A, B)
    I = @index(Global)
    @inbounds C[I] = A[I] + B[I]
end

function main(device)
  @show device
  ka_add! = ka_knl_add!(device, device isa CPU ? 4 : 1024)
  for k = 2:2:28
    N = 2^k

    A = rand(N)
    B = rand(N)
    C = similar(A)
    if device isa CUDADevice
      A = CuArray(A)
      B = CuArray(B)
      C = CuArray(C)
    end

    wait(ka_add!(C, A, B, ndrange = N))
    if C ≠ A + B
      println()
      println("2^$k = $N")
      @show C == A + B
      @show C ≈ A + B
      @show extrema(Array(C - A - B))
    end
  end
end

Random.seed!(77)
main(CPU())
main(CUDADevice())
```

But for some arrays sizes on a V100 this produces:

```
device = CPU()

device = CUDADevice()

2^16 = 65536
C == A + B = false
C ≈ A + B = false
extrema(Array((C - A) - B)) = (-5.447940142435828, 0.13998980843263142)

2^18 = 262144
C == A + B = false
C ≈ A + B = false
extrema(Array((C - A) - B)) = (-1.9034060006220799, 1.8951141473846453)
```

vchuravy:
Fixes #221, by placing synchronizing against the task local stream for kernels that are not
launched with dependencies. Maybe we should generally synchronize KA launches against the
task-local stream?

Also uses the task-local stream instead of the `CuDefaultStream`.

### DIFF
--- a/lib/CUDAKernels/src/CUDAKernels.jl
+++ b/lib/CUDAKernels/src/CUDAKernels.jl
@@ -186,7 +186,7 @@ function threads_to_workgroupsize(threads, ndrange)
     end
 end
 
-function (obj::Kernel{CUDADevice})(args...; ndrange=nothing, dependencies=nothing, workgroupsize=nothing, progress=yield)
+function (obj::Kernel{CUDADevice})(args...; ndrange=nothing, dependencies=Event(CUDADevice()), workgroupsize=nothing, progress=yield)
 
     ndrange, workgroupsize, iterspace, dynamic = launch_config(obj, ndrange, workgroupsize)
     # this might not be the final context, since we may tune the workgroupsize
--- a/test/test.jl
+++ b/test/test.jl
@@ -221,7 +221,7 @@ if backend != CPU
         event1 = kernel_empty(backend(), 1)(ndrange=1)
         event2 = kernel_empty(backend(), 1)(ndrange=0; dependencies=event1)
         @test event2 == MultiEvent(event1)
-        event = kernel_empty(backend(), 1)(ndrange=0)
+        event = kernel_empty(backend(), 1)(ndrange=0, dependencies=nothing)
         @test event == MultiEvent(nothing)
     end
 end


lib/CUDAKernels/src/CUDAKernels.jl
```diff
@@ -184,7 +184,7 @@ function threads_to_workgroupsize(threads, ndrange)
     end
 end
 
-function (obj::Kernel{CUDADevice})(args...; ndrange=nothing, dependencies=nothing, workgroupsize=nothing, progress=yield)
+function (obj::Kernel{CUDADevice})(args...; ndrange=nothing, dependencies=Event(CUDA()), workgroupsize=nothing, progress=yield)
```
mwarusz, line 187:
```suggestion
function (obj::Kernel{CUDADevice})(args...; ndrange=nothing, dependencies=Event(CUDADevice()), workgroupsize=nothing, progress=yield)
```

lib/CUDAKernels/src/CUDAKernels.jl
```diff
@@ -65,7 +65,7 @@ failed(::CudaEvent) = false
 isdone(ev::CudaEvent) = CUDA.query(ev.event)
 
 function Event(::CUDADevice)
-    stream = CUDA.CuDefaultStream()
+    stream = CUDA.stream()
```
jkozdon, line 68: With @mwarusz fix I am getting an error that `stream` is not defined.

vchuravy:
Tim pointed out that it needs ***@***.***`.

vchuravy:
bors try

vchuravy:
bors try

PR #230, opened 2021-03-30
Fix Bors

DilumAluthge:
Script used to generate the list of statuses: https://github.com/JuliaGPU/KernelAbstractions.jl/blob/master/.ci/bin/get-github-status-list.jl

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -2,9 +2,10 @@ delete_merged_branches = true
 
 status = [
   ### GPU CI (Buildkite):
-  "buildkite/kernelabstractions-dot-jl/julia-v1",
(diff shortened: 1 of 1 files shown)


DilumAluthge:
bors mrge

DilumAluthge:
bors merge

DilumAluthge:
bors merge

jpsamaroo:
I suspect AMDGPU has a race during kernel launch that causes us to sometimes not write out kernel packets correctly; usually restarting CI fixes it (as you can see from the now-passing ROCM tests). I'll file an issue for it.

PR #236, opened 2021-04-02
Add new overdub for unsafe_getindex to avoid allocating error message in Julia 1.6+

jakebolewski:

PR #245, opened 2021-04-29
Fix UndefVarError

michel2323:
* Fixed UndefVarError introduced in https://github.com/JuliaGPU/KernelAbstractions.jl/pull/236

Moving from `v0.6.0` to `v0.6.1` triggered a few undefined variable errors. This PR fixes those.


vchuravy:
Can you also add a test-case?

michel2323:
@vchuravy Added an `:` operator test, but not sure in the right place.

michel2323:
Okay, works with the CPU but not with the CUDAKernels backend. I guess `:` should not be used this way.

test/colon.jl
```diff
@@ -0,0 +1,15 @@
+using KernelAbstractions, Test
+
+@kernel function copy_colon_kernel!(A, @Const(B))
+    I = @index(Global)
+    @inbounds A[I] = sum(B[:,I])
```
vchuravy, line 5:
```suggestion
    @inbounds A[I] = sum(view(B, :,I))
```
maybe?

michel2323: That solves our problem in `ExaPF`, but then doesn't trigger the `unsafe_getindex` overdub in https://github.com/JuliaGPU/KernelAbstractions.jl/blob/51a59dc8ff2427b53db93a2e50efb4a972e370f5/src/compiler.jl#L94 .

michel2323: https://github.com/exanauts/ExaPF.jl/blob/34f3a7103851f04c61b578808880ad64aeba739d/src/LinearSolvers/preconditioners.jl#L277  this is another statement that breaks for the same reason but without `:`.

michel2323:
@vchuravy I couldn't get an MWE where `_unsafe_getindex` is called other than with the colon, although we get it triggered in ExaPF. However, I could not get why. So here is a test that directly calls `_unsafe_getindex`.

michel2323:
Bad bot.

michel2323:
@vchuravy I guess you have to decide whether `_unsafe_getindex` should be supported. As of now [this code ](https://github.com/JuliaGPU/KernelAbstractions.jl/blob/51a59dc8ff2427b53db93a2e50efb4a972e370f5/src/compiler.jl#L94-L100) only works on the CPU backend. @jakebolewski What's your take?

jakebolewski:
I would revert that PR, I could never get it to work correctly (which is why I couldnt write the test and shelved it a while back) and we are going in a different direction with some of our project's kernels so I don't think it is useful in the near term.

vchuravy:
Fixed by  #246

PR #249, opened 2021-05-11
Stop overdubbing CUDA math functions

ali-ramadhan:
I think with CUDA.jl v3 there is no longer a need to overdub/rewrite CUDA specific math functions.

If anything `CUDA.pow` doesn't exist anymore.

### DIFF
--- a/test/compiler.jl
+++ b/test/compiler.jl
@@ -12,6 +12,10 @@ end
     A[1] = 2^11
 end
 
+@kernel function square(A, B)
+    A[1] = B[1]^2
+end
+
 @kernel function checked(A, a, b)
     A[1] = Base.Checked.checked_add(a, b)
 end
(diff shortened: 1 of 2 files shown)


vchuravy:
As I mentioned on slack, things are a bit more complicated since we don't want to overdub these, but we still have to use Cassette for the time being.

vchuravy:
Can you also add a test case for pow?

vchuravy:
bors try

test/compiler.jl
```diff
@@ -28,6 +32,11 @@ function compiler_testsuite()
         @test !any(stmt->(stmt isa Expr) && stmt.head == :invoke, CI.code)
     end
 
+    let (CI, rt) = @ka_code_typed square(CPU())(zeros(1), zeros(1), ndrange=1)
```
vchuravy, line 35: I noticed to late that this is just `CPU()` and thus doesn't check the CUDAKernels implementation

PR #251, opened 2021-05-13
Backport #249

vchuravy:
Backport #249.

@ali-ramadhan I noticed that you only tested `CPU()` execution,
but the changes are limited to CUDAKernels.

PR #252, opened 2021-05-13
Fix CUDA pow test

ali-ramadhan:
X-Ref: https://github.com/JuliaGPU/KernelAbstractions.jl/pull/251#issue-644226534


vchuravy:
ah we can't just do that :) you need to pass in the backend in from the external side of things

```
function compiler_testsuite(backend)
```

and then use `backend` in your test.

vchuravy:
Fixed by #253

PR #256, opened 2021-05-29
overdub erfcx

CarloLucibello:
follow-up to #115


vchuravy:
With #288 we no longer provide separate functions. These now need to go into CUDA.jl and AMDGPU.jl respectivly

PR #266, opened 2021-10-06
Add AMD section to a couple of examples

ali-ramadhan:
If AMDGPU.jl and ROCKernels.jl is mature enough then is it worth adding some AMD examples to the KernelAbstractions.jl examples?

This PR just tacks on an AMD example to the simplest example, `memcopy.jl`.

The `has_rocm_gpu` util doesn't belong here obviously (https://github.com/JuliaGPU/AMDGPU.jl/issues/158).

I guess there's some room for improvement since `mycopy!` is defined almost identically three times, but maybe being verbose here makes for a clearer example.

Hmmm, if AMD examples cannot be tested yet that would be a problem though.

### DIFF
--- a/examples/memcopy.jl
+++ b/examples/memcopy.jl
@@ -37,14 +37,6 @@ if has_cuda_gpu()
     @test A == B
 end
 
-function has_rocm_gpu()
-    for agent in AMDGPU.get_agents()
-        if agent.type == :gpu
-            return true
-        end
-    end
-    return false
-end
 
 if has_rocm_gpu()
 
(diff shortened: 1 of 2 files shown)


vchuravy:
bors try

ali-ramadhan:
Latest version of AMDGPU.jl exports `has_rocm_gpu` so I removed the local definitions from this PR.

vchuravy:
bors try

vchuravy:
Can you update the compat bound in `ROCKernels`?

Issue #265, opened 2021-10-02
Convert related kernels errors with CUDAKernels


mloubout:
kernels involving `convert` error due to call to julia runtime in the GPU case (runs fine on CPU). Here is an MFE

```
using CUDA, KernelAbstractions, CUDAKernels

@kernel function myceil(x, y)
    I = @index(Global)
    @inbounds x[I] = ceil(Int, y[I])
end

@kernel function basiconvert(x, y)
    I = @index(Global)
    @inbounds x[I] = convert(Int, y[I])
end

ker_ceil = myceil(CUDADevice(), 8)
ker_convert = basiconvert(CUDADevice(), 8)

x = CUDA.ones(Int, 111)
y = CUDA.randn(111)
z = cu(Vector{Float32}(1:111))
ker_ceil(x, y, ndrange=length(x))
ker_convert(x, z, ndrange=length(x))
```

Versions:
julia : 1.6.1
KernelAbstractions: 0.7.0
CUDA : 3.4.2
CUDAKernels : 0.3.0

mloubout:
Maybe duplicate of #254

vchuravy:
Should be fixed on 0.7.1

Issue #254, opened 2021-05-19
Compilation error with type-converting round functions, e.g., ceil(Int32, 1.2f0)


cdsousa:
Some time ago, CUDA.jl won't work with calls to rounding functions that perform conversion to integer types.
Currently, it already works (I don't know why since those versions are still not defined https://github.com/JuliaGPU/CUDA.jl/blob/4c9a03d6202afe816707ccd12ed22acfa65a3df5/src/device/intrinsics/math.jl#L243)

However, KernelAbstractions.jl still doesn't work, but instead it gives a seemingly unrelated error:

```julia
using CUDA
using KernelAbstractions

x_d = CUDA.cu(rand(Float32, 1))
y_d = CUDA.zeros(Int32, 1)

f(x::Float32) = ceil(Int32, x)

## CUDA

function cu!(y, x)
    y[1] = f(x[1])
    return nothing
end

y_d .= 0
@cuda cu!(y_d, x_d)
y_d

## KA

function main(y_d)
    @kernel function ka!(y, x)
        y[1] = f(x[1]) 
    end
    event = ka!(CUDAKernels.CUDADevice(), 1)(y_d, x_d, ndrange=(1,))
    wait(event)
end

y_d .= 0
main(y_d)
y_d
```

vchuravy:
Should be fixed on 0.7.1

PR #272, opened 2021-10-20
fixing shmem line to match CUDA.jl

leios:
This should fix any shared memory issues with KernelAbstractions; however, in the testing script, I had a convert from `Int64` -> `Int32` that caught #265 and potentially #254. I will post some findings there as well


vchuravy:
Can we lowerbound the `Project.toml`?

PR #283, opened 2022-01-11
Simplify CI and drop bors

vchuravy:


vchuravy:
Okay we are hanging on 1.7

PR #291, opened 2022-02-28
Use Timer to avoid hangs due to error on the device

### LINKED ISSUE #290: `wait(kernel(...)` performs much worse on Julia v1.7
Sometimes `wait(kernel(...))` hangs up.
MWE:

```julia
using KernelAbstractions
using CUDA, CUDAKernels
using ProgressMeter

@kernel function test_kernel(a::AbstractArray{T}) where {T}
    index = @index(Global)
    a[index] = T(1)
end

a = CuArray(zeros(Float32, 500, 500));
ndrange = size(a)

k = test_kernel(CUDADevice())
wait(k(a; ndrange))

@showprogress for i in 1:1000000 # get stuck at some point
    wait(k(a; ndrange))
end
```

I observed this on Julia v1.7.1 and v1.7.2.
On Julia v1.6.5, however, it works and never gets stuck. 
Package versions:
```julia
(@v1.7) pkg> st
      Status `~/.julia/linux-ubuntu-20.04-x86_64/environments/v1.7/Project.toml`
  [052768ef] CUDA v3.8.3
  [72cfdca4] CUDAKernels v0.3.3
  [63c18a36] KernelAbstractions v0.7.2
  [92933f4c] ProgressMeter v1.7.1

```

I tested it also with different machines (Ubuntu and Windows) and different GPUs (all NVIDIA's).

As @maleadt suggested this might be related to https://github.com/JuliaGPU/CUDA.jl/issues/1350 and https://github.com/JuliaGPU/CUDA.jl/pull/1369.

vchuravy:
Mimicks https://github.com/JuliaGPU/CUDA.jl/pull/1369 and may fix #290

### DIFF
--- a/lib/CUDAKernels/src/CUDAKernels.jl
+++ b/lib/CUDAKernels/src/CUDAKernels.jl
@@ -265,7 +265,7 @@ end
 
 import CUDA: @device_override
 
-import KernelAbstractions: CompilerMetadata, CompilerPass, DynamicCheck, LinearIndices
+import KernelAbstractions: CompilerMetadata, DynamicCheck, LinearIndices
 import KernelAbstractions: __index_Local_Linear, __index_Group_Linear, __index_Global_Linear, __index_Local_Cartesian, __index_Group_Cartesian, __index_Global_Cartesian, __validindex, __print
 import KernelAbstractions: mkcontext, expand, __iterspace, __ndrange, __dynamic_checkbounds
 
--- a/lib/CUDAKernels/src/CUDAKernels.jl
+++ b/lib/CUDAKernels/src/CUDAKernels.jl
@@ -76,13 +76,37 @@ wait(ev::CudaEvent, progress=yield) = wait(CPU(), ev, progress)
 function wait(::CPU, ev::CudaEvent, progress=nothing)
     isdone(ev) && return nothing
 
-    event = Base.Threads.Event()
+    event = Base.Event()
     stream = next_stream()
     wait(CUDADevice(), ev, nothing, stream)
     CUDA.launch(;stream) do
         notify(event)
     end
-    wait(event)
+    dev = CUDA.device()
+    # if an error occurs, the callback may never fire, so use a timer to detect such cases
+    timer = Timer(0; interval=1)
+    Base.@sync begin
+        Threads.@spawn try
+            device!(dev)
+            while true
+                try
+                    Base.wait(timer)
+                catch err
+                    err isa EOFError && break
+                    rethrow()
+                end
+                if CUDA.unsafe_cuStreamQuery(stream) != ERROR_NOT_READY
+                    break
+                end
+            end
+        finally
+            notify(event)
+        end
+        Threads.@spawn begin
+            Base.wait(event)
+            close(timer)
+        end
+    end
 end
 
 # Use this to synchronize between computation using the task local stream
--- a/lib/CUDAKernels/src/CUDAKernels.jl
+++ b/lib/CUDAKernels/src/CUDAKernels.jl
@@ -76,6 +76,21 @@ wait(ev::CudaEvent, progress=yield) = wait(CPU(), ev, progress)
 function wait(::CPU, ev::CudaEvent, progress=nothing)
     isdone(ev) && return nothing
 
+    # minimize latency of short operations by busy-waiting,
+    # initially without even yielding to other tasks
+    spins = 0
+    while spins < 256
+        if spins < 32
+            ccall(:jl_cpu_pause, Cvoid, ())
+            # Temporary solution before we have gc transition support in codegen.
+            ccall(:jl_gc_safepoint, Cvoid, ())
+        else
+            yield()
+        end
+        isdone(ev) && return
+        spins += 1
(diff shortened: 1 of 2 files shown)


lib/CUDAKernels/src/CUDAKernels.jl
```diff
@@ -76,13 +76,37 @@ wait(ev::CudaEvent, progress=yield) = wait(CPU(), ev, progress)
...
     CUDA.launch(;stream) do
         notify(event)
     end
-    wait(event)
+    dev = CUDA.device()
+    # if an error occurs, the callback may never fire, so use a timer to detect such cases
+    timer = Timer(0; interval=1)
+    Base.@sync begin
+        Threads.@spawn try
+            device!(dev)
```
lmh91, line 90: I guess this should be `CUDA.device!(dev)`?

lib/CUDAKernels/src/CUDAKernels.jl
```diff
@@ -76,13 +76,37 @@ wait(ev::CudaEvent, progress=yield) = wait(CPU(), ev, progress)
...
+            while true
+                try
+                    Base.wait(timer)
+                catch err
+                    err isa EOFError && break
+                    rethrow()
+                end
+                if CUDA.unsafe_cuStreamQuery(stream) != ERROR_NOT_READY
```
lmh91, line 98: And `CUDA.ERROR_NOT_READY`?

lmh91 (commented):
After changing these two lines it works and the issue https://github.com/JuliaGPU/KernelAbstractions.jl/issues/290 seems to be fixed. 

However, it is a massive slow down in performance.
About a factor of 10 in my use case. 
The slow down of the dummy kernel, which I posted in the issue, is even worse.

lmh91:
I copied and adapted https://github.com/JuliaGPU/CUDA.jl/blob/9d69cab5b6ccde944c0d2d7040b3787192c2ac14/lib/cudadrv/stream.jl#L141-L154
into CUDAKerners.jl and now the performance is much better again.

My program is a field solver via Succesive Over-Relaxation on a three dimensional grid. 
For smaller grid sizes I am now actually gaining some speed boost (about 2x).
However, for larger grids it seems I still get a worse performance (about 0.5x) compared to the current release version of CUDAKernels.jl and Julia v1.6.5. So not as bad as without the spinning but still observable.

lib/CUDAKernels/src/CUDAKernels.jl
```diff
@@ -76,13 +76,52 @@ wait(ev::CudaEvent, progress=yield) = wait(CPU(), ev, progress)
...
+        Threads.@spawn try
+            CUDA.device!(dev)
+            while true
+                try
+                    Base.wait(timer)
+                catch err
+                    err isa EOFError && break
+                    rethrow()
+                end
+                if CUDA.unsafe_cuEventQuery(ev.event) != CUDA.ERROR_NOT_READY
```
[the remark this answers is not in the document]
maleadt: Unsure, and I couldn't quickly test because most kernel errors (making a stream fail) are fatal anyway.

The strange thing is that this didn't happen on 1.6, and users aren't reporting any CUDA errors, so I'm not sure if it's related to the stream failing...

vchuravy:
Yeah nothing is failing, but we are still deadlocking...

lmh91:
Just wanted to confirm from my use case that the issues seems to be solved with the current version of this PR :relaxed:
But there is still some performance loss :neutral_face:

## Julia v1.7.1 & this PR:
```julia
BenchmarkTools.Trial: 807 samples with 1 evaluation.
 Range (min … max):  6.120 ms …   9.696 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     6.184 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   6.193 ms ± 144.802 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%
              
  ▂▁▃▂▂▁▂▃▃▃▃▄▄▄▄▅▆▆▆████████████████▆▅▆▆▄▂▃▃▃▃▃▂▃▂▂▃▁▃▂▂▂▂▂▂ ▄
  6.12 ms         Histogram: frequency by time        6.26 ms <

 Memory estimate: 39.78 KiB, allocs estimate: 512.
```

## Julia 1.6.5 & latest releases:
```
BenchmarkTools.Trial: 1386 samples with 1 evaluation.
 Range (min … max):  3.478 ms …   7.129 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.592 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.604 ms ± 131.082 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%
                        
  ▂▂▁▂▁▁▂▁▃▂▃▃▃▄▆█████████████████▇▆▆▄▅▄▄▃▃▄▃▂▃▂▃▂▂▂▁▂▂▂▁▁▂▁▂ ▄
  3.48 ms         Histogram: frequency by time        3.76 ms <

 Memory estimate: 76.09 KiB, allocs estimate: 676.
```

PR #297, opened 2022-05-22
Support Adapt.AbstractGPUDevice

oschulz:
See https://github.com/JuliaGPU/Adapt.jl/pull/52

Seems functional for CUDA.jl, at least ("examples/matmul.jl" runs). Will certainly need expert input.


oschulz:
Closed, but not forgotten, to be reincarnated in the future (see JuliaGPU/Adapt.jl#64).

PR #288, opened 2022-02-09
Excise Cassette

vchuravy:

PR #300, opened 2022-05-25
using CPU as default for Scratchpad and SharedMemory

leios:
I have been getting this error ever since #288 

```
[ Info: Precompiling KernelAbstractions [63c18a36-062a-441e-b654-da1e3ab1ce7c]
WARNING: Method definition SharedMemory(Type{T}, Base.Val{Dims}, Base.Val{x} where x) where {T, Dims} in module KernelAbstractions at /home/leios/projects/KernelAbstractions.jl/src/KernelAbstractions.jl:464 overwritten at /home/leios/projects/KernelAbstractions.jl/src/cpu.jl:204.
  ** incremental compilation may be fatally broken for this module **

WARNING: Method definition Scratchpad(Any, Type{T}, Base.Val{Dims}) where {T, Dims} in module KernelAbstractions at /home/leios/projects/KernelAbstractions.jl/src/KernelAbstractions.jl:460 overwritten at /home/leios/projects/KernelAbstractions.jl/src/cpu.jl:218.
  ** incremental compilation may be fatally broken for this module **
```

I think it's because there used to be an overdub for ScratchPad and SharedMemory on the CPU with a default to an error message in KernelAbstractions, but now we are using `@device_override` and using the CPU versions as defaults.

This PR just moves the stuff in `cpu.jl` to `KernelAbstractions.jl`. There might be a better solution, so it's up to you what you want to do.

### DIFF
--- a/src/KernelAbstractions.jl
+++ b/src/KernelAbstractions.jl
@@ -457,12 +457,22 @@ include("macros.jl")
 # Backends/Interface
 ###
 
-function Scratchpad(ctx, ::Type{T}, ::Val{Dims}) where {T, Dims}
-    throw(MethodError(Scratchpad, (T, Val(Dims))))
+@inline function SharedMemory(::Type{T}, ::Val{Dims}, ::Val) where {T, Dims}
+    MArray{__size(Dims), T}(undef)
 end
 
-function SharedMemory(::Type{T}, ::Val{Dims}, ::Val{Id}) where {T, Dims, Id}
-    throw(MethodError(SharedMemory, (T, Val(Dims), Val(Id))))
+###
+# CPU implementation of scratch memory
+# - memory allocated as a MArray with size `Dims`
+###
+
+struct ScratchArray{N, D}
+    data::D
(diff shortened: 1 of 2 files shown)


src/KernelAbstractions.jl
```diff
@@ -457,12 +457,22 @@ include("macros.jl")
 # Backends/Interface
 ###
 
-function Scratchpad(ctx, ::Type{T}, ::Val{Dims}) where {T, Dims}
-    throw(MethodError(Scratchpad, (T, Val(Dims))))
+@inline function SharedMemory(::Type{T}, ::Val{Dims}, ::Val) where {T, Dims}
```
vchuravy, line 460: Can we leave the implementation back to CPU.jl and here just define:

```
function Scratchpad end
```

vchuravy:
Thanks! That's a silly oversight of me

src/KernelAbstractions.jl
leios: I changed it and also added `function SharedMemory end`.

vchuravy (approved):

PR #305, opened 2022-06-08
Enable CPU atomics

jpsamaroo:


jpsamaroo:
That error relates to operations on floating point values, which will be partially addressed by https://github.com/JuliaConcurrent/UnsafeAtomics.jl/pull/9

PR #311, opened 2022-06-25
bounding UnsafeAtomics and UnsafeAtomicsLLVM

leios:
This should fix #305 so users to not need to manually add UnsafeAtomics separately for CPU atomics and also https://github.com/JuliaRegistries/General/pull/62061#issuecomment-1165914744


Project.toml
```diff
@@ -12,12 +12,14 @@ MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
+UnsafeAtomics = "013be700-e6cd-48c3-b4a1-df204f14c38f"
```
vchuravy, line 15: @leios I think you also need to say `using UnsafeAtomics` somewhere otherwise the code isn't loaded

vchuravy (approved):

PR #312, opened 2022-06-28
Add oneAPI back-end.

maleadt:
TODO:

- [ ] Don't use Float64 in tests; the oneAPI CI GPU doesn't support it
- [ ] Events: do we need to signal the event when creating it? Other back-ends do it, but that seems fishy.
- [ ] wait: do we need to support an invocation with no queue argument (having it default to the global queue)? again, other back-ends support this, but it just seems wrong to use the global queue (which is generally not used by KA.jl) for signalling an event.


lib/oneAPIKernels/src/oneAPIKernels.jl
```diff
@@ -0,0 +1,283 @@
...
+failed(::oneAPIEvent) = false
+isdone(ev::oneAPIEvent) = Base.isdone(ev.event)
+
+function Event(::oneAPIDevice)
+    ctx = oneAPI.context()
+    dev = oneAPI.device()
+    event_pool = oneAPI.ZeEventPool(ctx, 1, dev)
+    # XXX: this is inefficient
+    event = event_pool[1]
+    # XXX: other back-ends record/signal here straight away. why?
```
vchuravy, line 76: The goal here is to have a Event that can be used to synchronize against the default stream.

E.g. if I used GPUArrays I will need to create an event to depend on, to make sure that KA runs after those kernelrs

maleadt: Ah, so it needs to be a globally-synchronized event for interop with other task-local packages.

vchuravy:
> wait: do we need to support an invocation with no queue argument (having it default to the global queue)?

Yes this is meant to synchronize the global stream so that no kernel on it launches before all of the KA tasks have finished. We need a bottom and a top synchronization for switching between KA code and GPUArrays code

maleadt:
OK, this passes tests. Any good stress test to verify the back-end?

vchuravy:
@glwagner or @michel2323 have complicated KA codes

michel2323:
Oh yes. And I will test that ASAP @maleadt . Probably today, maybe tomorrow.

michel2323:
Just a minor fix was required to get it off the ground.
```
diff --git a/lib/oneAPIKernels/src/oneAPIKernels.jl b/lib/oneAPIKernels/src/oneAPIKernels.jl
index 11f39ce..e51e41d 100644
--- a/lib/oneAPIKernels/src/oneAPIKernels.jl
+++ b/lib/oneAPIKernels/src/oneAPIKernels.jl
@@ -85,8 +85,8 @@ end
 function Event(::oneAPIDevice)
     event = new_event()
     queue = default_queue()
-    execute!(queue) do list
-        append_signal!(list, event)
+    oneL0.execute!(queue) do list
+        oneL0.append_signal!(list, event)
     end
     oneAPIEvent(event)
 end
```
However, then it crashes in this kernel. https://github.com/exanauts/ExaAdmm.jl/blob/c600c7618f6b22a0b0fac4d2c473061e2ac3e4c4/src/models/acopf/acopf_admm_update_x_ka.jl#L20 defined here https://github.com/exanauts/ExaAdmm.jl/blob/ms/oneapi/src/models/acopf/acopf_auglag_linelimit_kernel_ka.jl

I used Julia 1.8-rc1.
```
  [052768ef] CUDA v3.11.0
  [72cfdca4] CUDAKernels v0.4.2
  [4d6a948c] ExaAdmm v0.2.0 `..`
  [63c18a36] KernelAbstractions v0.8.3 `~/.julia/dev/KernelAbstractions`
  [8f75cd03] oneAPI v0.2.4
  [3b98bdbd] oneAPIKernels v0.1.0 `~/.julia/dev/KernelAbstractions/lib/oneAPIKernels`
  [4af54fe1] LazyArtifacts
  [37e2e46d] LinearAlgebra
  [de0858da] Printf
  [fa267f1f] TOML v1.0.0
  [8dfed614] Test
```
Logs are attached, but I guess they won't be very useful. I'll try to build an MWE. All I can say is that these tests pass with AMDGPU.jl and CUDA.jl. At the time of the crash, some kernels have already finished before. This is done on an integrated GPU. I'll check this out on another machine.

[dbg.log](https://github.com/JuliaGPU/KernelAbstractions.jl/files/9004777/dbg.log)

maleadt:
Oh, it crashes the driver, that's interesting. No need to reduce this, first try dumping `compiled.image` here: https://github.com/JuliaGPU/oneAPI.jl/blob/51f12496270c5d17cab76228950e338391edb97d/src/compiler/execution.jl#L161= (just write it to disk), and send it to me.

maleadt:
`InvalidBitWidth: Invalid bit width in input: 128`
I guess there's an int128 somewhere.

maleadt:
Should be good to go.

Coverage change unrelated -- it's because we now don't decrypt secrets anymore for pull requests, and Codecov doesn't support tokenless upload for Buildkite. I've adapted our config so that the token does get decrypted on non-third party PRs, but the coverage check should probably still be relaxed (sadly).

vchuravy:
Fantastic!

PR #316, opened 2022-07-25
ROCKernels: Update to AMDGPU 0.4

jpsamaroo:


vchuravy:
Needs to update the Project.toml?

jpsamaroo:
> Needs to update the Project.toml?

Somehow it missed the push...

So now we have a `ROCDevice` in AMDGPU.jl. What should we do with `ROCKernels.ROCDevice`? Any ideas for a better name?

jpsamaroo:
I switched to `KAROCDevice` in the meantime. Let me know if anyone has better name ideas :smile:

jpsamaroo:
What's up with the MPI failures?

vchuravy:
`KAROCDevice` reads absolutly horrid. You also need to update ROCKernels to 0.4 since this is a breaking change.  Maybe `ROCmDevice`? Can we cheat and just re-use `ROCDevice` from `AMDGPU.jl`

jpsamaroo:
> Maybe ROCmDevice? Can we cheat and just re-use ROCDevice from AMDGPU.jl

But it would need to be `<: GPU`, no?

vchuravy:
I don't think we use that dispatch rule for much...

vchuravy:
Can you remove the Manidest.toml you added?

jpsamaroo:
@vchuravy the `<: GPU` is necessary to get `<: Device` dispatches. Would it be reasonable to manually forward these for `ROCDevice`?

vchuravy:
Sure!

jpsamaroo:
Ok, now all calls generated by the `@kernel` macro forward down to `KernelAbstractions.construct`, which is defined for `::Device`, and ROCKernels defines its own for `::ROCDevice`.

leios (approved):
It seems to run for me locally. The MPI errors seem to be a general CI problem right now?

jpsamaroo:
Yeah, the MPI failures seem to be something specific to CI.

Let's hold off on merging this until:
- We've tested that this works properly for CUDAKernels (since we change some internals to support using `AMDGPU.ROCDevice`
- We've added the `ROCDevice()` ctor to AMDGPU, so that we can keep doing that (it's much nicer for users)

PR #318, opened 2022-09-07
updating ROCKernels to AMDGPU v0.4.1

leios:
This should update ROCKernels to the latest AMDGPU.

This is necessary because otherwise AMDGPU is pinning CUDA at v 3.9.1 and causing the following build error: https://github.com/JuliaGPU/CUDA.jl/pull/1558

I realized after the update that `ROCDevice` is now potentially conflicting with `AMDGPU.ROCDevice`.


leios:
I am an idiot and somehow missed #316. Sorry for the spam!

jpsamaroo:
Closing in favor of #316, as the necessary changes are slightly more involved.

PR #327, opened 2022-11-09
fix typo in docs

bjarthur:


bjarthur:
relatedly, in [Quickstart / Terminology](https://github.com/JuliaGPU/KernelAbstractions.jl/blob/master/docs/src/quickstart.md?plain=1#L9-L11) it says "If using a workgroup of size 1 (non-parallel execution), the ndrange is the number of items to iterate over in a loop.", but judging by how `ndrange` is used in the examples, isn't it *always* equal to the number or items, independent of the workgroup size?

PR #329, opened 2022-11-15
Add @ndrange

vchuravy:


leios:
So quick question on this:

```
@kernel function f_test_kernel!(input)
    @print(@ndrange(), '\n')
end
```

with `input = zeros(2,2)` outputs `CartesianIndex{1}[CartesianIndex(1,), CartesianIndex(2,), CartesianIndex(3,), CartesianIndex(4,)]` per thread.  I guess I would have expected it instead to just output what I set as `ndrange` (in this case `length(input) == 4`).

It's no huge deal because I can just do `size(@ndrange)` to get what I want, but should the `Tuple{Int}` output be default?

Issue #330, opened 2022-11-15
Simple Block Reduce Fails when using `while` loops


anicusan:
Hi, thank you for developing this library - I would like to write optimised kernels for common GPU algorithms such as reduce, scan, radix sort, etc. similar to [CUB](https://nvlabs.github.io/cub/) but available on all KernelAbstractions platforms. The resulting "KA standard library" (KALib? Caleb?) could be used as a benchmark for future KA development & optimisation - and I can use the lessons along the way to populate the "Writing Kernels" section in the documentation. Big plans, but...

I'm implementing the block-wise reduce following [this tutorial](https://developer.download.nvidia.com/assets/cuda/files/reduction.pdf) with this simple-looking code:

```julia
using KernelAbstractions
using CUDA
using CUDAKernels


@kernel function block_reduce(out, in)

    # Get block / workgroup size
    bs = @uniform @groupsize()[1]

    # Block / group index, thread index within block, global thread index
    bi = @index(Group, Linear)
    ti = @index(Local, Linear)
    gi = @index(Global, Linear)

    # Copy each thread's corresponding item from global to shared memory
    cache = @localmem eltype(out) (bs,)
    cache[ti] = in[gi]
    @synchronize

    # Reduce elements in shared memory using sequential addressing following
    # https://developer.download.nvidia.com/assets/cuda/files/reduction.pdf
    @private s = bs ÷ 2
    while s > 0
        if ti < s
            cache[ti] += cache[ti + s]
        end

        @synchronize

        s = s >> 1
    end

    # Copy result back to global memory
    if ti == 1
        out[bi] = cache[1]
    end
end


num_blocks = 10
block_size = 32
num_elements = num_blocks * block_size

in = rand(1:10, num_elements) |> CuArray
out = zeros(num_blocks) |> CuArray


kernel_reduce = block_reduce(CUDADevice(), block_size)
ev = kernel_reduce(out, in, ndrange=num_elements)

wait(ev)
println(out)

```

It shouldn't be more exotic than the example code in the docs - however, these two lines:

```julia
    @private s = bs ÷ 2
    while s > 0
```

Produce the following errors:

```julia
Reason: unsupported use of an undefined name (use of 'bs')
Stacktrace:
 [1] macro expansion
   @ ~/Prog/Julia/KALib/prototype/reduce.jl:31
 [2] gpu_block_reduce
   @ ~/.julia/packages/KernelAbstractions/DqITC/src/macros.jl:81
[...]

PR #335, opened 2022-11-16
Fix quickstart

vchuravy:
On the CPU the workgroupsize is not a factor of the number of threads, but rather a multiply of the vector width.

### DIFF
--- a/docs/src/quickstart.md
+++ b/docs/src/quickstart.md
@@ -5,7 +5,7 @@ Because CUDA is the most popular GPU programming environment, we can use it as a
 reference for defining terminology in KA. A *workgroup* is called a block in
 NVIDIA CUDA and designates a group of threads acting in parallel, preferably
 in lockstep. For the GPU, the workgroup size is typically around 256, while for the CPU,
-it is usually a multiply of the natural vector-width. An *ndrange* is
+it is usually a multiple of the natural vector-width. An *ndrange* is
 called a grid in NVIDIA CUDA and designates the total number of work items. If
 using a workgroup of size 1 (non-parallel execution), the ndrange is the
 number of items to iterate over in a loop.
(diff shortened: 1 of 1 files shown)


docs/src/quickstart.md
```diff
@@ -5,7 +5,7 @@ Because CUDA is the most popular GPU programming environment, we can use it as a
 reference for defining terminology in KA. A *workgroup* is called a block in
 NVIDIA CUDA and designates a group of threads acting in parallel, preferably
 in lockstep. For the GPU, the workgroup size is typically around 256, while for the CPU,
-it is usually less than or equal to the number of CPU cores. An *ndrange* is
+it is usually a multiply of the natural vector-width. An *ndrange* is
```
leios, line 8:
```suggestion
it is usually a multiple of the natural vector-width. An *ndrange* is
```
This is actually news to me, almost all the examples use 256 threads and _4 cores_. We should change those.

But stupid questiona(s): why don't we use cores? Also: what is a vector-width?

vchuravy: I don't know where the 16 snuck in from. We default to 1024 when none is specified.

https://github.com/JuliaGPU/KernelAbstractions.jl/blob/732fc8450637899db7be318e43b36ac05321fa33/src/cpu.jl#L104

It's actually the opposite way around length(ndrange)/length(wkgrp) should be approx nthreads (or multiple thereof since you want to allow for some work balancing)

vchuravy: Generally speaking each workgroup is processed by a single task, so you want to balance the overhead of creating tasks.

This tradeoff might change with https://github.com/JuliaGPU/KernelAbstractions.jl/issues/330#issuecomment-1317625382 but I am not sure how to implement that transform yet

PR #337, opened 2022-11-21
Add Metal support

tgymnich:
fixes #326

PR #338, opened 2022-11-21
Metal CI

tgymnich:


vchuravy:
@maleadt pointed out that the Metal CI isn't sandboxed https://github.com/JuliaGPU/KernelAbstractions.jl/pull/337#issuecomment-1322345902 so we would need to go back to bors.

maleadt:
Metal.jl uses `bors`, but I'm not sure if we can extend that to KernelAbstractions.jl without impacting other back-ends.

PR #340, opened 2022-12-13
Add zeroinit argument to @localmem

michel2323:


vchuravy:
Can you also add a test? And I think you will need to actually implement it for the other backends

jpsamaroo:
The zeroinit logic in AMDGPU is pretty generic: it just finds GVs which have a `__zeroinit_` prefix and then adds a `memset` call for each at the top of the entry function; so it could easily be upstreamed to GPUCompiler, if desired.

jpsamaroo:
Seems like CUDA shmem is already zeroinit by default, so the CUDAKernels implementation should probably just ignore it.

vchuravy:
I still don't understand the rational behind this PR. What's the concrete goal here?

vchuravy:
I think the right answer is that init is required. CPU defaults to `undef` and for everything else is "implementation defined"

vchuravy:
https://github.com/JuliaGPU/KernelAbstractions.jl/blob/f3e7b1cf3d10c2970dd74c9945c9054150abab55/src/cpu.jl#L221

> I am not sure that is user-friendly without an explicit undef

Oh I agree. The issue is that we don't know what zero is for all types. So we should definitely document this better

michel2323:
I would also make it explicit in the code. If one sees `undef` alarm bells go off.

jpsamaroo:
I'm personally OK with forcing zeroing behavior for all backends (regardless of whether it's sensible for the element type or not), although zeroinit for ROCKernels will imply some performance cost, as we have to synchronize the wavefront after zeroinit completion.

luraess:
Zero init has significant performance costs in uses cases such as caching stencil operation into shmem. Having a same strategy across all backends is crucial for portability. Ideally, default behavior could be non init as this would be the performance oriented use case.

vchuravy:
Closing this as stale.

PR #343, opened 2023-01-30
[CUDAKernels] add always_inline as device parameter

vchuravy:


vchuravy:
Hm in retro-spect I am not sure why I added these as type-parameters...

We never pass the device to the device code. So normal parameters should have been fine.

lcw:
Having the  parameter tied to the device kinda makes sense as this only works on the CUDA backend.  For example, it seems that AMDGPU has turned on inlining by default <https://github.com/JuliaGPU/AMDGPU.jl/commit/2e52d1d53e246b1ccfdc4d1b6b734473eec526cd>.

lib/CUDAKernels/src/CUDAKernels.jl
```diff
@@ -95,8 +95,8 @@ end
 
 import KernelAbstractions: Event, CPUEvent, NoneEvent, MultiEvent, CPU, GPU, isdone, failed
 
-struct CUDADevice{PreferBlocks} <: GPU end
-CUDADevice() = CUDADevice{false}()
+struct CUDADevice{PreferBlocks, AlwaysInline} <: GPU end
+CUDADevice(;prefer_blocks=false, always_inline=false) = CUDADevice{prefer_blocks, always_inline}()
```
vchuravy, line 99:
```suggestion
CUDADevice(;prefer_blocks=false, always_inline=false) = CUDADevice{prefer_blocks, always_inline}()
CUDADevice{PreferBlocks}() where PreferBlocks = CUDADevice{PreferBlocks, false}()
```

PR #351, opened 2023-02-07
Add infrastructure for EnzymeRules

vchuravy:
Still need to decide if we should use Requires instead of hard includ for pre-1.9

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,12 +14,20 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 UnsafeAtomics = "013be700-e6cd-48c3-b4a1-df204f14c38f"
 UnsafeAtomicsLLVM = "d80eeb9a-aca5-4d75-85e5-170c8b632249"
+EnzymeCore = "f151be2c-9106-41f4-ab19-57ee4f262869"
 
 [compat]
 Adapt = "0.4, 1.0, 2.0, 3.0"
 Atomix = "0.1"
 MacroTools = "0.5"
 StaticArrays = "0.12, 1.0"
-UnsafeAtomicsLLVM = "0.1"
 UnsafeAtomics = "0.2.1"
+UnsafeAtomicsLLVM = "0.1"
+EnzymeCore = "0.2"
 julia = "1.7"
+
+[weakdeps]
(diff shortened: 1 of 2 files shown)


christophernhill:
@vchuravy and @wsmoses I tried a little with this and the Oceananigans mini test at 

https://github.com/christophernhill/dj4earth-onan-noodling/tree/cnh/test_noka_enz0.10/tests 

I compared a non-KA case

   https://github.com/christophernhill/dj4earth-onan-noodling/tree/cnh/test_noka_enz0.10/tests/test02
   
and a KA with rules ( https://github.com/christophernhill/KernelAbstractions.jl/tree/vc/rules_cnh_fix_Project ) and with slightly modified kernel functions in Onan ( https://github.com/christophernhill/OceanLES.jl/tree/cnh/onan_ka_enz ). 

Three interesting things. Item (1.) looks to me like maybe something weird is going on in arguments somewhere? (2.) and (3.) are more just to remind me some other clean up is needed at some stage - not really tied to this PR. (1.) seems like something is a bit wonky? 

 1. the adjoint sensitivity plot here  ( https://github.com/christophernhill/dj4earth-onan-noodling/blob/cnh/test_noka_enz0.10/tests/test03/using-ka.png )
the KA with rules case looks wrong. It looks like some sort of difference between the initial and final value of the independent variable ( c ) and not the partial derivative. The non KA case looks like this ( 
 https://github.com/christophernhill/dj4earth-onan-noodling/blob/cnh/test_noka_enz0.10/tests/test02/no-ka-pre-changes.png
 ) which is closer to what it should be. 
 
 2. I ended up messing with https://github.com/christophernhill/KernelAbstractions.jl/blob/vc/rules_cnh_fix_Project/Project.toml there was a circular Onan dependency and also the `[weakdep]` kept causing problems. 
 
 3. In ( 
 https://github.com/christophernhill/dj4earth-onan-noodling/blob/cnh/test_noka_enz0.10/tests/test02/no-ka-pre-changes.png
 )  something looks a tiny bit off in the periodic _halo_ bit I think. I am thinking about that, maybe there is a reason for that I am overlooking. 
 
 I started some notes that can make this a bit of a Literate tutorial and test case when it works. 
[enzyme_diff_example.pdf](https://github.com/JuliaGPU/KernelAbstractions.jl/files/10788815/enzyme_diff_example.pdf) - notes are in tex for now.

PR #360, opened 2023-03-05
Let Event(MtlDevice) actually be a barrier

vchuravy:


lib/MetalKernels/src/MetalKernels.jl
```diff
@@ -74,7 +74,16 @@ failed(::MetalEvent) = false
...
 function Event(::MetalDevice)
-    event = Metal.MtlSharedEvent(Metal.current_device())  ### FIXME: Event vs SharedEvent
+    device = Metal.current_device()
+    queue = Metal.global_queue(device)
+    cmdbuf = Metal.MtlCommandBuffer(queue)
+    event = Metal.MtlSharedEvent(device)
+
+    # The cmdbuf executes in-order on the global queue
+    # so this one just triggers our signal
+    MTL.encode_signal!(cmdbuf, event, METAL_EVENT_SIGNAL_VALUE)
```
vchuravy, line 84: @maleadt Do I need an `Metal.enqueue!(buf)` before this?

https://github.com/JuliaGPU/Metal.jl/blob/fdab27774c5f4b96ad7fc8bafee698394dde8e70/src/state.jl#L53`

tgymnich: no https://developer.apple.com/documentation/metal/mtliocommandbuffer/3951185-enqueue?language=objc

tgymnich:
You might also wanna set the default dependencies for a kernel launch to MetalEvenet() to mirror the default behavior in CUDAKernels.

PR #363, opened 2023-03-08
Add optional priority feature

vchuravy:


vchuravy:
@luraess is this roughly what you had in mind? To give the communication task a higher priority?

examples/mpi.jl
```diff
@@ -21,7 +21,8 @@ end
 
 function exchange!(h_send_buf, d_recv_buf, h_recv_buf, src_rank, dst_rank, comm)
     recv_req = MPI.Irecv!(h_recv_buf, src_rank, 666, comm)
-    recv = Base.Threads.@spawn begin 
+    recv = Base.Threads.@spawn begin
+        KernelAbstractions.priority!(backend, :high)
         cooperative_test!(recv_req)
         KernelAbstractions.copyto!(backend, d_recv_buf, h_recv_buf)
         KernelAbstractions.synchronize(backend) # Gurantueed to be cooperative
```
luraess, line 28:
```suggestion
        KernelAbstractions.synchronize(backend) # Guaranteed to be cooperative
```

luraess:
Yeah - that's the direction. How is it implemented now? Do you have a pool of queues / streams per priority?

vchuravy:
> Do you have a pool of queues / streams per priority?

No this would change the priority of the stream owned by the Task you are running on.

luraess:
> No this would change the priority of the stream owned by the Task you are running on.

True - need to get used to the task based approach. Now with priorities, can one then implement a workflow to have one kernel to execute parts of the computations on the high priority queue / stream, have the MPI comm on that same task (I guess), and compute inner points on another low priority stream (asynchronously on another task)?

PR #353, opened 2023-02-09
Add adapt storage rules for device

utkarsh530:
Fixes TODO in https://github.com/SciML/DiffEqGPU.jl/pull/227/

PR #364, opened 2023-03-09
Backends are adaptors

vchuravy:


vchuravy:
Replaces #353

@maleadt good idea or bad idea?

src/KernelAbstractions.jl
```diff
@@ -360,6 +360,11 @@ get_backend(A::Tridiagonal) = get_backend(A.d)
 
 get_backend(::Array) = CPU()
 
+# Define:
+#   adapt_storage(::Backend, a::Array) = adapt(BackendArray, a)
+#   adapt_storage(::Backend, a::BackendArray) = a
+Adapt.adapt_storage(::CPU, a::Array) = a
```
maleadt, line 366: Why the instance and not the type?

vchuravy: You mean the `::Type{CPU}`< I thought we defined adaptors over the instance. The user might not have the backend type around, but they should always have an instance.

maleadt: well, for CuArray we do `adapt(CuArray, obj)`, i.e., specifying the type because that's likely the only thing you have at that point. If users don't explicitly mention the backend types I guess it would be fine to use the instance here.

PR #369, opened 2023-03-21
Add queries for atomics and float64 support

maxwindiff:


src/KernelAbstractions.jl
```diff
@@ -386,6 +386,20 @@ function ones(backend, ::Type{T}, dims::Tuple) where T
     return data
 end
 
+"""
+    is_atomics_suppported(::Backend)
+
+Returns whether `@atomic` operations are supported by the backend.
+"""
+is_atomics_suppported(backend) = true
```
maxwindiff, line 394: @vchuravy Is this naming convention ok? And do we want to export these?

maxwindiff: Done, updated!

PR #379, opened 2023-04-05
Introducing @reduce for group level reduction

brabreda:
The @reduce macro performs a group level reduction. 

TODOs:
- Figure out a place for the implementation.
- Add a lane level reduction.
- Create a more advanced group level reduction that is able to utilize platform dependant feature such as lane reduction and  atomics.

### DIFF
--- a/lib/CUDAKernels/src/CUDAKernels.jl
+++ b/lib/CUDAKernels/src/CUDAKernels.jl
@@ -323,7 +323,6 @@ import CUDA: @device_override
 import KernelAbstractions: CompilerMetadata, DynamicCheck, LinearIndices
 import KernelAbstractions: __index_Local_Linear, __index_Group_Linear, __index_Global_Linear, __index_Local_Cartesian, __index_Group_Cartesian, __index_Global_Cartesian, __validindex, __print
 import KernelAbstractions: mkcontext, expand, __iterspace, __ndrange, __dynamic_checkbounds
-import KernelAbstractions: __reduce
 
 
 function mkcontext(kernel::Kernel{<:CUDADevice}, _ndrange, iterspace)
@@ -400,6 +399,14 @@ end
     CUDA._cuprint(args...)
 end
 
+import KernelAbstractions: __test
+
+@device_override @inline function __test(__ctx__, conf) 
+    KernelAbstractions.@localmem Float64 conf.threads_per_block
+
+    KernelAbstractions.@print("dit werkt")
+end
+
 ###
 # GPU implementation of const memory
 ###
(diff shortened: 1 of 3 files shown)


lib/CUDAKernels/src/CUDAKernels.jl
```diff
@@ -407,4 +409,40 @@ Adapt.adapt_storage(to::ConstAdaptor, a::CUDA.CuDeviceArray) = Base.Experimental
...
 
+# TODO: make variable block size possible
+# TODO: figure out where to place this
+# reduction functionality for a group
+@device_override @inline function __reduce(__ctx__ , op, val, neutral, ::Type{T}) where T
+    threads = KernelAbstractions.@groupsize()[1]
+    threadIdx = KernelAbstractions.@index(Local)
+
+    # shared mem for a complete reduction
+    shared = KernelAbstractions.@localmem(T, 1024)
```
vchuravy, line 420: Maybe this is the moment we need dynamic shared memory support?

src/KernelAbstractions.jl
```diff
@@ -329,6 +330,14 @@ macro index(locale, args...)
     Expr(:call, GlobalRef(KernelAbstractions, index_function), esc(:__ctx__), map(esc, args)...)
 end
 
+# TODO: Where should we havdle the logic of neutral, adding it to the macro's logic would reduce complexity in terms of using the macro
+# but adding it to the macro may cause some overhead
+macro reduce(op, val, neutral)
```
vchuravy, line 335: Is the question here if we should auto-detect the neutral element for op + Type?

I think this is fine, expecting the user to provide their own neutral element.

lib/CUDAKernels/src/CUDAKernels.jl
```diff
@@ -407,4 +409,40 @@ Adapt.adapt_storage(to::ConstAdaptor, a::CUDA.CuDeviceArray) = Base.Experimental
...
+    @inbounds shared[threadIdx] = val
+
+    # perform the reduction
+    d = 1
+    while d < threads
+        KernelAbstractions.@synchronize()
```
vchuravy, line 426: You are inside `CUDAKernels` here and as such you can use CUDA.jl functionality directly.

vchuravy:
Looks like a great start! Will have to add it to `0.9` but that can happen after you are happy with the initial implementation.

lib/CUDAKernels/src/CUDAKernels.jl
vchuravy: x-ref: #11

lib/CUDAKernels/src/CUDAKernels.jl
brabreda: Thats correct! But a implementation with KA.jl macros would allow for a single implementation that can run on all supported back-end. Because of this I am not sure what the best place is for the code for this implementation. 

Also, the main difference between different back-end would the size of local memory but the use of dynamic memory would be a solution to this.

src/KernelAbstractions.jl
brabreda: Yes, that was the question.

brabreda:
To make a more generalized @reduce operation, I would work with a Config struct. An example of this can be found in the [GemmKernels.jl Config](https://github.com/JuliaGPU/GemmKernels.jl/blob/590d9e81b4cf1653667c771ae56285c8c717ce42/src/config.jl#L1-L37).

Based on this struct, the reduction could use atomics and lane/warp reductions.

PR #381, opened 2023-04-10
Add unsafe_free!

pxl-th:


src/KernelAbstractions.jl
```diff
@@ -85,6 +85,17 @@ Synchronize the current backend.
...
 
+"""
+    unsafe_free!(x::AbstractArray)
+
+Release the memory of an array for reuse by future allocations.
+This function is automatically called by the finalizer
+when an array goes out of scope, but can be called
+earlier to reduce pressure on the memory allocator.
+
+On CPU backend this is a no-op.
```
vchuravy, line 96: Maybe reword this without the reference to the finalizer, since this method in particular will not be used.

Note that implementing it is optional (not only on the CPU) and what the contract is for the user.

PR #382, opened 2023-04-13
Enzyme Rules, all functioning

wsmoses:
Requires https://github.com/EnzymeAD/Enzyme.jl/pull/729/checks 

I failed to understand the extension work, so please feel free to rework.


test/testsuite.jl
```diff
@@ -17,6 +17,7 @@ end
 
 include("test.jl")
 include("localmem.jl")
+include("enzyme.jl")
```
vchuravy, line 20: You forgotten to check in the testfile

vchuravy:
Also will this only work for the CPU? I would prefer if we could have a generalized implementation that would also work for the GPU backends.

wsmoses:
@vchuravy forward mode is functioning with the rule in CUDA, etc.

reverse mode, is not because of the previously discussed tape issues. I've actually mostly worked around it, however that would need to be in CUDAKernels.jl as a rule registration to be able to get the parent job (And also because mkcontext, etc is apparently not stable from cpu -> gpu etc).

vchuravy:
Fails in function verification:

```
Call parameter type does not match function signature!
{} addrspace(10)* ()* bitcast ({ { {} addrspace(10)*, {} addrspace(10)*, {} addrspace(10)* } ()*, void (i8*)* }* @"_enzyme_reverse_julia_wait_3789'" to {} addrspace(10)* ()*)
 {} addrspace(10)*  %16 = call {} addrspace(10)* @julia_enq_work_3818({} addrspace(10)* ()* bitcast ({ { {} addrspace(10)*, {} addrspace(10)*, {} addrspace(10)* } ()*, void (i8*)* }* @"_enzyme_reverse_julia_wait_3789'" to {} addrspace(10)* ()*)), !dbg !3333
```

vchuravy:
Ah that's just 1.6 being unsupported

wsmoses:
@vchuravy excludijf nightly which is currently not supported by Enzyme per the swiftcc stuff, this fully builds and runs successfully on CI.

can we merge?

vchuravy:
@maleadt Reverse CI for CUDA seems to be broken?

Maybe https://github.com/JuliaGPU/CUDA.jl/pull/1970?

vchuravy:
@wsmoses did you test this with CUDA?

wsmoses:
I did not, but only forward mode has a rule for all backends. Otherwise the PR intentionally just adds reverse mode for CPU. Reason is the tape nested job stuff we discussed, which will take some thinking to solve.

test/enzyme.jl
```diff
@@ -0,0 +1,33 @@
...
+end
+
+function enzyme_testsuite(backend, ArrayT)
+    @testset "kernels" begin
+        A = ArrayT{Float64}(undef, 64)
+        A .= (1:1:64)
+        dA = ArrayT{Float64}(undef, 64)
+        dA .= 1
+
+        Enzyme.autodiff(Reverse, caller, Duplicated(A, dA), Const(backend()))
```
vchuravy, line 23: So this test would not pass on GPU backends?

We will need to do feature detection then.

wsmoses: Yeah

Issue #387, opened 2023-04-25
`CPU(static=true)` option


carstenbauer:
I think it would be valuable to add a `CPU(static=true)` option to the `CPU` backend that runs a kernel with `@threads :static` instead of `@spawn`. `CPU(static=false)` could still be the default such that this would be non-breaking.

On multi-NUMA domain systems it is crucial to make sure that pages get established on the correct NUMA node. Pragmatically, this may be realized by first-touch policy, i.e. initializing the data in parallel via `@threads :static` ([example](https://github.com/JuliaPerf/STREAMBenchmark.jl/blob/master/src/benchmarks.jl#L38-L61)). However, ideally, the kernel must also use the same threads to access the same pieces of the data to run efficiently. Because KA.jl is currently using `@spawn`, it is non-deterministic on which thread tasks might end up running on and consequently one gets subpar performance (even worse, it can fluctuate heavily).

(cc @vchuravy)

vchuravy:
I will trade you implement this for a good example/documentation on how to use KA with NUMA aware code.

vchuravy:
> If such an example plus a description of what is going on / why this is necessary is good enough, we have a deal wink

Yes!

Yeah we could make these https://github.com/JuliaGPU/KernelAbstractions.jl/blob/edaa0242dbedc3f57f46729aacb7546a1ac6a33d/src/cpu.jl#L7-L8 use `static=true`.

carstenbauer:
Xref: https://github.com/JuliaGPU/KernelAbstractions.jl/pull/388

navidcy:
Was this resolved by #388?

PR #388, opened 2023-04-25
Add CPU(static=true)

vchuravy:


src/cpu.jl
```diff
@@ -16,7 +16,7 @@ function (obj::Kernel{CPU})(args...; ndrange=nothing, workgroupsize=nothing, )
         return nothing
     end
 
-    __run(obj, ndrange, iterspace, args, dynamic)
+    __run(obj, ndrange, iterspace, args, dynamic, kernel.backend.static)
```
carstenbauer, line 19:
```suggestion
    __run(obj, ndrange, iterspace, args, dynamic, obj.backend.static)
```

src/KernelAbstractions.jl
```diff
@@ -353,7 +353,10 @@ constify(arg) = adapt(ConstAdaptor(), arg)
 abstract type Backend end
 abstract type GPU <: Backend end
 
-struct CPU <: Backend end
+struct CPU <: Backend
+    static::Bool
+    CPU(;static::Bool=false) = CPU(static)
```
carstenbauer, line 358:
```suggestion
    CPU(;static::Bool=false) = new(static)
```

carstenbauer (commented):
With my suggested changes, this gives

```julia
julia> measure_membw(CPU());
        Memory Bandwidth (GB/s): 182.01
        Compute (GFLOP/s): 30.33

julia> measure_membw(CPU(; static=true));
        Memory Bandwidth (GB/s): 345.3
        Compute (GFLOP/s): 57.55
```
for the SAXPY example.

carstenbauer:
Not sure if you want to add it here in this PR but anyways:

```julia
allocate(::CPU, ::Type{T}, dims::Tuple) where T = Array{T}(undef, dims)
function _parallel_init(backend::CPU, f::F, ::Type{T}, dims::Tuple) where {F, T}
    println("parallel_init")
    arr = allocate(backend, T, dims)
    Threads.@threads :static for i in eachindex(arr)
        @inbounds arr[i] = f(T)
    end
    return arr
end
function zeros(backend::CPU, ::Type{T}, dims::Tuple) where T
    if backend.static
        _parallel_init(backend, zero, T, dims)
    else
        Base.zeros(T, dims)
    end
end
function ones(backend::CPU, ::Type{T}, dims::Tuple) where T
    if backend.static
        _parallel_init(backend, one, T, dims)
    else
        Base.ones(T, dims)
    end
end
```

With this one can initialize e.g. with `rand!(KernelAbstractions.zeros(backend, Float64, N))` to get the performance mentioned above.

carstenbauer:
Xref: https://github.com/JuliaGPU/KernelAbstractions.jl/issues/387

PR #399, opened 2023-06-12
Attempt at including offsets in kernel launch

simone-silvestri:
This PR tries to include offsets in kernel launches so that the `Global` indices returned by 
`@index(Global, NTuple)` and `@index(Global, Linear)` are offset by an `offset` argument.

Example:
```julia
julia> @kernel function show_index()
             i, j = @index(Global, NTuple)
             @show i, j
       end
show_index (generic function with 6 methods)

julia> show_index(CPU(), (2, 2), (3, 3), (-1, -2))()
(i, j) = (0, -1)
(i, j) = (1, -1)
(i, j) = (0, 0)
(i, j) = (1, 0)
(i, j) = (2, -1)
(i, j) = (2, 0)
(i, j) = (0, 1)
(i, j) = (1, 1)
(i, j) = (2, 1)
```
where the last argument `(-1, -2)` is the offsets to the global indices.

This PR constrains the offsetting of global indices on static kernel size at launch.

@vchuravy I found it a bit difficult to implement arbitrary indices because of the division in blocks, which would have to be rethought. Aka, this is the easiest (probably not the most general) implementation of offsets. Let me know if you would rather it be implemented in another way.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -6,6 +6,7 @@ steps:
           - "1.6"
           - "1.7"
           - "1.8"
+          - "1.9"
     plugins:
       - JuliaCI/julia#v1:
           version: "{{matrix.version}}"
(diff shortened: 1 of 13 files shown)


vchuravy:
Thanks for the initial implementation I will have to think about this a bit.
I still feel like this may be better expressed as a projection `f(Idx) -> Idx`.

```
abstract type Projection end
struct Identity <: Projection end
(::Identity)(Idx) = Idx

struct Offset{Offsets} <: Projection end
```

vchuravy:
@timholy might also be able to offer advise. IIUC you are trying to implement an exterior/interior iteration split like `EdgeIterator` from https://github.com/JuliaArrays/TiledIteration.jl?

luraess:
@vchuravy following this as ability to handle ranges passed to kernels is also a feature that we would necessitate (FD MPI code) to allow for communication computation overlap (in a similar way as pointed out by @simone-silvestri).

vchuravy:
You can do this right now as you would do with CUDA.jl/AMDGPU.jl by projecting a smaller ndrange to your custom index space. This is more about if we can do something like that automatically.

@lcw I think had some code that does this for his DG code

luraess:
Yeah - having something more automatised could be a nice thing. @utkinis may have a small MWE on what we did recently which would be handy to have as well in KA (similar to the proposed thing).

PR #401, opened 2023-06-16
Add implementation notes to host functionality

### LINKED ISSUE #400: KA seems to be broken for CUDA
Whenever I try to call a function that should use my GPU (GeForce RTX 4090) I either get an overflow error or a never-ending loop. For example: 
```
dev = CUDA.device()
KernelAbstractions.zeros(dev, Float32, 10, 5) 
```    
or 
```
get_backend(cu(rand(1)))
```
.

vchuravy:
Specifies what is and what isn't mandated functionality.

Fixes #400


maleadt:
Didn't realize this also included the fix. LGTM!

PR #415, opened 2023-08-22
enable dependabot for GitHub actions

ranocha:
This allows to get updates for GitHub actions automatically. I have used this for my own packages, the [Trixi.jl framework](https://github.com/trixi-framework), and the [SciML organization](https://github.com/SciML). After merging this, you could also enable other Dependabot actions in 'Settings -> Code security and analysis -> Dependabot alerts' and '... -> Dependabot security updates'.

See https://github.com/SciML/MuladdMacro.jl/pull/37

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "monthly"


vchuravy:
I don't really see the value in automated tooling like this? What does this get me?

ranocha:
Some of the actions that you're using are deprecated. For example, I see

> The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

in some of your latest CI runs such as https://github.com/JuliaGPU/KernelAbstractions.jl/actions/runs/5640024406

I don't know how GitHub will handle this in the future. In their blog, they write

> What you need to do
> 
> For Actions maintainers: Update your actions to run on Node16 instead of Node12 ([Actions configuration settings](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runs-for-javascript-actions))
> For Actions users: Update your workflows with latest versions of the actions which runs on Node16 ([Using versions for Actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-versioned-actions))

Dependabot will do the latter for you (by creating a PR)

vchuravy:
*sigh* I do so enjoy churn.

PR #419, opened 2023-09-01
Config struct + reductions

brabreda:
This pull request introduces:
- a config object that can be used to use specific features depending on a backend.
- reduction API

Another idea would be to include ndrange in this config object. With this, we would be able to base the kernellaunch on this object so no need to define groupsize and ndrange again.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,0 +1,261 @@
+# This file is machine-generated - editing it directly is not advised
+
+julia_version = "1.8.5"
+manifest_format = "2.0"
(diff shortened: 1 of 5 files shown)


vchuravy:
Thank you, this is interesting. A couple quick comments:

1. The dependency relationship between GPUArrays and KernelAbstractions should be inverted. E.g. KA should be a dependency of GPUArrays and used there to implement map reduce.
2. The new functionality requires tests and documentation.
3. How does Config differ from Kernel? Could we merge them into the same steuct?

brabreda:
Okay! I will fix this!

It centainly has alot of simiulair

Issue #262, opened 2021-07-20
`@synchronize`  inside while loops doesn't work on CPUs


tom91136:
I'm trying to port the classic GPU tree reduction via KernelAbstractions.jl.
See [this](https://github.com/UoB-HPC/BabelStream/blob/6fe81e19556ac26761a1c7247ae29fa88fb4e0ab/CUDAStream.cu#L233) for the direct CUDA implementation of what I'm trying to port from.
This is what I have implemented currently:
```julia

const TBSize = 1024::Int
const DotBlocks = 256::Int

@kernel function dot(@Const(a), @Const(b), size, partial)  
    local_i = @index(Local)
    group_i = @index(Group)
    tb_sum = @localmem T TBSize
    @inbounds tb_sum[local_i] = 0.0
  
    # do dot first
    i = @index(Global)
    while i <= size
      @inbounds tb_sum[local_i] += a[i] * b[i]
      i += TBSize * DotBlocks
    end
  
    # then tree reduction
    offset = @private Int64 (1,)
    @inbounds begin
      offset[1] = @groupsize()[1] ÷ 2
      while offset[1] > 0
        @synchronize
        if (local_i - 1) < offset[1]
          tb_sum[local_i] += tb_sum[local_i+offset[1]]
        end
        offset[1] ÷= 2
      end
    end
  
    if (local_i == 1)
      @inbounds partial[group_i] = tb_sum[local_i]
    end
  end

# driver
wait(dot(backendDevice, TBSize)(a, b, size, partial_sum, ndrange = TBSize * DotBlocks))
```
I was able to get correct results and performance seems mostly on par with our CUDA.jl(https://github.com/UoB-HPC/BabelStream/blob/7c1e04a42b9b03b0e5c5d0b07c0ef9f4bdd59353/JuliaStream.jl/src/CUDAStream.jl#L112) and AMDGPU.jl(https://github.com/UoB-HPC/BabelStream/blob/7c1e04a42b9b03b0e5c5d0b07c0ef9f4bdd59353/JuliaStream.jl/src/AMDGPUStream.jl#L135) implementation.
On CPU however, I got the following error:

```
Using device: AMD Ryzen 9 3900X 12-Core Processor (1T)
ERROR: LoadError: TaskFailedException
Stacktrace:
 [1] wait
   @ ./task.jl:322 [inlined]
 [2] wait
[...]

PR #421, opened 2023-09-09
groupreduction and subgroupreduction

brabreda:
I am unsure why my previous PR closed but here are the changes.
- I added docs
- I added tests

It was my first time writing tests, and they passed. How are these tested on GPUs, and do I need more tests?


vchuravy:
Thank you! As you found the notion of a subgroup is not necessarily consistent across backends. So I am wondering if we need to expose that as feature or if backends could use them to implement reduction.

So as an example could we "just" add `rval= @reduce(op, val)` to perform a workgroup level reduction?

What is `rval` across all members of a workgroup. Is it the same or may it divergent and only the first workitem will receive a correct answer?

Do we need a broadcast/shuffle operation? 

I don't have the answers to those questions. 

I think adding `@reduce` is definitely something we want to do, but subgroup ops may need more thought.

brabreda:
- `rval=@reduce(op, val)` is possible. In this case we only use `localmem` and `@synchronize` for the reduction. 

- at the moment only the first thread of the group returns an item, but I think it would be possible to change this.

- I Think providing a neutral value will have better performance, but I am not sure. I Will check this. 

I will fix this, meanwhile we can think about the wraps. If we will introduce it and in what way.

vchuravy:
Yeah requiring a neutral value sounds good.

Could not each backend choose to use warp level ops, iff available? Instead of having the user make this decision.

brabreda:
The problem with that approach is that warp intrinsics don’t work for every type. 

We then would need something like this:

`__reduce(op, val::union{type1, …},)` for wraps
`__reduce(op, val,)` for everything else 

In this case you don’t override the function. I am also not sure if this kind of type specialisation works where it takes the first function when it is one of the types. You probs have more expertise on this.

would it be prefered that every thread returns the right rval, I do think in that case I Will have to rethink the warp reduction bases groupreduce.

vchuravy:
> would it be prefered that every thread returns the right rval? I do think in that case I Will have to rethink the warp reduction bases groupreduce.

No it just needs to be documented :) and we might want to add a `@broadcast`.


So it looks like CUDA.jl makes this decision based on the element type?
https://github.com/JuliaGPU/CUDA.jl/blob/d57e02018ccf00ceda7672f5f4d98e2ceef9106d/src/mapreduce.jl#L177-L181

src/KernelAbstractions.jl
```diff
@@ -139,10 +139,31 @@ function unsafe_free! end
...
+"""
+    @subgroupsize()
+
+    returns the GPUs subgroupsize.
+"""
+macro subgroupsize()
+    quote
+        $__subgroupsize()
+    end
+end
```
vchuravy, line 154:
```suggestion
```

src/reduce.jl
```diff
@@ -0,0 +1,98 @@
...
+    end
+
+    return val
+end
+
+@inline function __groupreduce(__ctx__, op, val, neutral, ::Type{T}, ::Val{false}) where {T}
+    idx_in_group = @index(Local)
+    groupsize = @groupsize()[1]
+    
+    localmem = @localmem(T, groupsize)
```
vchuravy, line 72: I see if we can do a subgroupreduce the memory we need here is much reduced.

src/reduce.jl
```diff
@@ -0,0 +1,98 @@
...
+
+"""
+
+@groupreduce(op, val, neutral, use_subgroups)
+
+Reduce values across a block
+- `op`: the operator of the reduction
+- `val`: value that each thread contibutes to the values that need to be reduced
+- `netral`: value of the operator, so that `op(netural, neutral) = neutral``
+- `use_subgroups`: make use of the subgroupreduction of the groupreduction
```
vchuravy, line 27: I see the value of having a common implementation.

So I would define:

```
@reduce(op, val, neutral)
__reduce(op, val, neutral)
```

And then maybe:

```
__subgroupreduce & __subgroupsize (No @ version)
__can_subgroup_reduce(T) = false
```

And then we could define:

```
__reduce(__ctx___, op, val, neutral, ::Val{true})
__reduce(__ctx___, op, val, neutral, ::Val{false})
```

as you have here.

```
function __reduce(op, val::T, neutral::T) where T
     __reduce(op, val, neutral, Val(__can_subgroup_reduce(T)))
end
```

brabreda:
Yeah so CUDA(and i think most GPU backends make the decision based on type). I was thinking if something like this would be possible:

For the implementation without warps:
`
@reduce(op,val,neutral)
__reduce(op,val,neutral)
`
And every back-end like CUDAKernels just implements something like this:
`
__reduce(op,val::T,neutral) where {T <: Union{Float32, Float64, etc...}}
`

I tried It today and I get a lot of dynamic invocation errors. But I am not sure what's at fault.

vchuravy:
I recommed `using Cthulhu, using CUDA` and then using `CUDA.device_code_typed interactive=true`

brabreda:
Only the group reduction without subgroups remains. It is ready to merge, once merged I will add support for Subgroups to the CUDAkernels, MetalKernels, and OneAPI.

src/reduce.jl
```diff
@@ -0,0 +1,49 @@
...
+    groupsize = @groupsize()[1]
+    
+    localmem = @localmem(T, groupsize)
+
+    @inbounds localmem[idx_in_group] = val
+
+    # perform the reduction
+    d = 1
+    while d < groupsize
+        @synchronize()
```
vchuravy, line 30: https://github.com/JuliaGPU/KernelAbstractions.jl/issues/262

brabreda:
The API is indeed neutral. But indeed the problem now is the @synchronize() in while loops.

src/reduce.jl
maleadt: Workaround?

PR #422, opened 2023-09-18
Add query function for if we are executing on a GPU

vchuravy:
Requires an API bump since backend need to implement this
with `device_overlay`.

### DIFF
--- a/src/KernelAbstractions.jl
+++ b/src/KernelAbstractions.jl
@@ -372,6 +372,13 @@ macro index(locale, args...)
     Expr(:call, GlobalRef(KernelAbstractions, index_function), esc(:__ctx__), map(esc, args)...)
 end
 
+"""
+    isongpu()
+
+Returns if this kernel is currently executing on a GPU.
+"""
+function isongpu end
+
 ###
 # Internal kernel functions
 ###
@@ -534,14 +541,6 @@ function priority!(::Backend, prio::Symbol)
     end
     return nothing
 end
-
-
-"""
-    ongpu()
-
-Returns if this kernel is currently executing on a GPU.
-"""
-function isongpu end
 
 include("nditeration.jl")
 using .NDIteration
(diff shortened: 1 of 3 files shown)


src/KernelAbstractions.jl
```diff
@@ -535,6 +535,14 @@ function priority!(::Backend, prio::Symbol)
     return nothing
 end
 
+
+"""
+    ongpu()
+
+Returns if this kernel is currently executing on a GPU.
+"""
+function ongpu end
```
simonbyrne, line 544: Prefix with `is` to make it easier to tab-complete?
```suggestion
function isongpu end
```

src/KernelAbstractions.jl
```diff
@@ -535,6 +535,14 @@ function priority!(::Backend, prio::Symbol)
     return nothing
 end
 
+
+"""
+    ongpu()
+
+Returns if this kernel is currently executing on a GPU.
+"""
+function isongpu end
```
MasonProtter, line 544: there's a mismatch here between the docstring and the function name. The method in `src/cpu` also calls in `ongpu`

maleadt:
Feels a little ad-hoc. I haven't thought this through, but what about a `Device` or `Accelerator` object, which could be used to query additional properties (SIMD width, etc)? The GPU-ness could be a part of the type hierarchy (`CUDAGPU <: AbstractGPU <: AbstractAccelerator` or whatever).

vchuravy:
Hm yeah, I was thinking `current_backend()` but Simon wanted the ability to not use it on the top-level.

We could pre-empt some other work I am thinking about and expose something like `target()` or `compilation_target()` and and then return a Features struct.

simonbyrne:
> We could pre-empt some other work I am thinking about and expose something like `target()` or `compilation_target()` and and then return a Features struct.

Any more thoughts on this?

PR #429, opened 2023-11-02
adding attempt to force inbounds at the kernel level

leios:
This adds an additional flag at the `@kernel` level to force `@inbounds` across the entire kernel. It works like:

```
julia> @kernel inbounds=true function check_inline(a)
           a[5] = 5
       end
julia> a = [0,0];
julia> backend = get_backend(a);
julia> kernel! = check_inline(backend, 256);
julia> kernel!(a, ndrange = 10) # no error!
```

Note that this comes with all the caveats of using `@inbounds` and is certainly not good for default behaviour, but... Geez I'm tired of all the `@inbounds`

I'll rebase and add some tests soon if people think this could be useful.

### DIFF
--- a/src/KernelAbstractions.jl
+++ b/src/KernelAbstractions.jl
@@ -53,28 +53,44 @@ synchronize(backend)
 ```
 """
 macro kernel(expr)
-    __kernel(expr, #=generate_cpu=#true)
+    __kernel(expr, #=generate_cpu=#true, #=force_inbounds=#false)
 end
 
 """
-    @kernel cpu=false function f(args) end
+    @kernel config function f(args) end
 
-Disable code-generation of the CPU function. This relaxes semantics such that
-KernelAbstractions primitives can be used in non-kernel functions.
+This allows for two different configurations:
+
+1. `cpu={true, false}`: Disables code-generation of the CPU function. This relaxes semantics such that KernelAbstractions primitives can be used in non-kernel functions.
+2. `inbounds={false, true}`: Enables a forced `@inbounds` macro around the function definition in the case the user is using too many `@inbounds` already in their kernel. Note that this can lead to incorrect results, crashes, etc and is fundamentally unsafe. Be careful!
 
 - [`@context`](@ref)
 
 !!! warn
     This is an experimental feature.
 """
-macro kernel(config, expr)
-    if config isa Expr && config.head == :(=) &&
-        config.args[1] == :cpu && config.args[2] isa Bool
-        generate_cpu = config.args[2]
+macro kernel(ex...)
+    if length(ex) == 1
+        __kernel(ex[1], true, false)
     else
-        error("Configuration should be of form `cpu=false` got $config")
+        generate_cpu = true
+        force_inbounds = false
+        for i = 1:length(ex)-1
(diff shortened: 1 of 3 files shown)


src/KernelAbstractions.jl
```diff
@@ -53,28 +53,44 @@ synchronize(backend)
 ```
 """
 macro kernel(expr)
-    __kernel(expr, #=generate_cpu=#true)
+    __kernel(expr, #=generate_cpu=#true, #=force_inline=#false)
```
vchuravy, line 56:
```suggestion
    __kernel(expr, #=generate_cpu=#true, #=force_inbounds=#false)
```

vchuravy (approved):

leios:
Ok, I think this PR is done. I made it kinda on a whim and I'm not sure if people actually want it. I could use it for a few specific workflows and think it might be useful for debugging https://github.com/JuliaGPU/AMDGPU.jl/issues/536

Maybe we can wait a few days to see if anyone else thinks it could be useful?

The way I see it, it's one of those things that makes the code more "kernel" like, but can also really hurt some people who are not careful.

vchuravy:
It's non breaking so I am happy to include it.

test/runtests.jl
```diff
@@ -21,6 +21,18 @@ kern_static(CPU(static=true), (1,))(A, ndrange=length(A))
...
+@kernel cpu=false inbounds=false function my_no_cpu_kernel2(a)
+end
+@test_throws ErrorException("This kernel is unavailable for backend CPU") my_no_cpu_kernel2(CPU())
+
+# testing bounds errors
+@kernel inbounds=false my_bounded_kernel(a) = a[1]
+@test_throws BoundsError(Int64[],(1,)) my_bounded_kernel(CPU())(Int[], ndrange=1)
+
+@kernel inbounds=true my_inbounds_kernel(a) = a[1]
+@test nothing == my_inbounds_kernel(CPU())(Int[], ndrange=1)
```
vchuravy, line 34: I think CI runs with `--check-bounds=true` which overrides `@inbounds`.

vchuravy (approved):

luraess:
> Ok, I think this PR is done. I made it kinda on a whim and I'm not sure if people actually want it. I could use it for a few specific workflows and think it might be useful for debugging https://github.com/JuliaGPU/AMDGPU.jl/issues/536
> 
> 
> 
> Maybe we can wait a few days to see if anyone else thinks it could be useful?
> 
> 
> 
> The way I see it, it's one of those things that makes the code more "kernel" like, but can also really hurt some people who are not careful.

That's nice to have! And can definitively help while debugging and developing.

lcw:
Love it!

leios:
Ok, looks like some other folks might find this useful. Also, after looking at https://github.com/JuliaGPU/AMDGPU.jl/issues/354, and https://github.com/JuliaLang/julia/issues/48245, I think this is a fair PR to merge.

Also `--check-bounds=yes` is set by default for `Pkg.test`. Does anyone know how to check whether this environmental variable is on or off in Julia itself? Otherwise, I am not sure how to test that the macro is actually successful.

leios:
Ok, I think the tests pass (except nightly). So I think the PR is good to go.

PR #431, opened 2023-11-06
Fix inbounds codegen for CPU

vchuravy:


leios (approved):
works locally and on tests

Issue #443, opened 2023-12-13
Allow `return` or `return nothing` in kernels


eschnett:
I use `JuliaFormatter` to format my source code. This adds explicit `return` statements. This leads to
```
ERROR: LoadError: Return statement not permitted in a kernel function sum2_kernel!
```
even if the `return` statement is just `return nothing`.

At the same time, ending a kernel with an expression (which is equivalent to a `return` statement) is allowed.

Could this check be changed to allow only empty `return` or `return nothing` statements?

vchuravy:
> At the same time, ending a kernel with an expression (which is equivalent to a return statement) is allowed.

KernelAbstractions will politly ignore those and insert a `return nothing` for you.
But yes we could explicitly allow `return` or `return nothing`

pxl-th:
Closed by #446.

PR #446, opened 2023-12-22
Allow empty return statements

eschnett:


eschnett:
Closes https://github.com/JuliaGPU/KernelAbstractions.jl/issues/443.

pxl-th:
Does it allow multiple return statements?

```julia
function ker()
    ...
    if something()
       return
    end
    ...
    if something_else()
       return
    end
    ...
    return
end
```

eschnett:
It does. But – should it?

pxl-th:
I'd like to have this option. Right now I find myself writing wrapper functions that I call from the kernel just to be able to use return statements.

eschnett:
*ping* Could you let the CI workflows run for this PR?

pxl-th (approved):

PR #481, opened 2024-06-10
Force Vararg specialization in CPU kernel launch

vchuravy:
Julia de-specialized varargs 

https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing

### DIFF
--- a/src/cpu.jl
+++ b/src/cpu.jl
@@ -34,7 +34,7 @@ function copyto!(backend::CPU, A, B)
     end
 end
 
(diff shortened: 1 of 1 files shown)


vchuravy:
@wsmoses IIUC this is no longer needed?

wsmoses:
yeah, at least OC no longer requires it to go through everything.

It may be useful to do for perf [or required for something else perhaps], but that's a bridge we can cross later.

PR #483, opened 2024-06-21
Enzyme: simplify via mixedduplicated

wsmoses:


ext/EnzymeExt.jl
```diff
@@ -150,7 +126,7 @@ module EnzymeExt
         args2 = ntuple(Val(N)) do i
             Base.@_inline_meta
             if args[i] isa Active
-                Duplicated(Ref(args[i].val), arg_refs[i])
+                EnzymeCore.MixedDuplicated(args[i].val, arg_refs[i])
```
vchuravy, line 129: This is correct for the CPU, but I think we won't be able to pass `MixedDuplicated` to the GPU. 
Or at least we need an Adapt rule.

wsmoses: Do we not have those already like we do for the other annotations?

wsmoses: Apparently not, now we do here: https://github.com/EnzymeAD/Enzyme.jl/pull/1551

wsmoses:
@vchuravy in interim could we get this merged/tagged (is blocking https://github.com/CliMA/Oceananigans.jl/pull/3618 / https://buildkite.com/clima/oceananigans/builds/16126#01903c25-aa5f-4e96-9088-70984b90ebe4 seemingly)

ext/EnzymeExt.jl
vchuravy: We should double check that this adapts to what we want on CUDA. I don't know if CuRef is right here...

wsmoses: Sure but this is code wouldn’t affect any currently landed cuda support (this is only reverse mode).

So we should likely explore that in michel’s PR?

luraess: And what about non CUDA GPUs?

vchuravy:
@wsmoses just so that you aware, this is what `Ref` get's converted to for `cudacall`.

https://github.com/JuliaGPU/CUDA.jl/blob/84bb117194d42c5ec58c09e85c312d491fedee68/src/compiler/execution.jl#L165-L175

PR #487, opened 2024-06-27
Fixing transpose function, a missing "end" and some argument orders.

Sixzero:
I believe things are correct this way.

### DIFF
--- a/examples/performance.jl
+++ b/examples/performance.jl
@@ -141,10 +141,11 @@ for block_dims in ((TILE_DIM, TILE_DIM), (TILE_DIM*TILE_DIM, 1), (1, TILE_DIM*TI
             output = similar(input)
 
             # compile kernel
-            kernel(input, output, ndrange=size(output))
+            kernel(output, input, ndrange=size(output))
             for rep in 1:nreps
-                KernelAbstractions.synchronize(backend)
+                kernel(output, input, ndrange=size(output))
             end
+            KernelAbstractions.synchronize(backend)
         end
     end
 end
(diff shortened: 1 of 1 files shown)


examples/performance.jl
```diff
@@ -143,7 +143,8 @@ for block_dims in ((TILE_DIM, TILE_DIM), (TILE_DIM*TILE_DIM, 1), (1, TILE_DIM*TI
             # compile kernel
             kernel(input, output, ndrange=size(output))
             for rep in 1:nreps
-            KernelAbstractions.synchronize(backend)
+                KernelAbstractions.synchronize(backend)
```
vchuravy, line 146: There should be a kernel launch before the sync

examples/performance.jl
```diff
@@ -161,7 +162,7 @@ for (name, kernel) in (
             # compile kernel
             kernel(input, output, Val(Int(bank)), ndrange=size(output))
             for rep in 1:nreps
-                kernel(input, output, Val(Int(bank)), ndrange=size(output))
+                kernel(output, input, Val(Int(bank)), ndrange=size(output))
```
vchuravy, line 165: This order needs to be changed everywhere

Sixzero: Oh, I forgot to change it everywhere... I guess because I just pushed it from GH editor... fixing it fast.

examples/performance.jl
Sixzero: Fixing it.

vchuravy (approved):

Sixzero:
How to rerun the checks? It failed with an Enzyme precompilation step.

vchuravy:
Thank you!

PR #500, opened 2024-08-05
Improve CPU launch heuristic

### LINKED ISSUE #499: Default CPU workgroupsize can be inadequat for higher-dimensionsal kernels
As observed in https://github.com/WaterLily-jl/WaterLily.jl/pull/133#issuecomment-2263500896

When we have a grid like `(64, 64, 64)` using a WG size of (1024, 1, 1) is inefficient. Instead WL uses `(64,1,1)` and maybe KA should "spread" the `1024` to `(64, 64,1)`.

vchuravy:
Fixes #499

cc: @b-fg @weymouth

### DIFF
--- a/src/cpu.jl
+++ b/src/cpu.jl
@@ -46,6 +46,29 @@ function (obj::Kernel{CPU})(args...; ndrange=nothing, workgroupsize=nothing, )
     __run(obj, ndrange, iterspace, args, dynamic, obj.backend.static)
 end
 
+const CPU_GRAINSIZE = 1024 # Vectorization, 4x unrolling, minimal grain size
+function default_cpu_workgroupsize(ndrange)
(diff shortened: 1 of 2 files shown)


vchuravy:
As expected performance improvement for small cases:

```
saxpy/default/Float32/4096 	3.03 ± 0.02 μs 	3.03 ± 0.02 μs 	1
saxpy/default/Float32/512 	1.27 ± 0.011 μs 	0.756 ± 0.11 μs 	1.68
saxpy/default/Float32/256 	1.22 ± 0.01 μs 	0.54 ± 0.12 μs 	2.26
saxpy/default/Float32/64 	1.19 ± 0.013 μs 	0.404 ± 0.0038 μs 	2.94
```

PR #505, opened 2024-08-09
Add runic as a formatter

vchuravy:


fredrikekre:
After your discourse comment I wanted to run Runic on this repo to check but was surprised the diff was empty. This PR explains it!

I scanned the diff here and didn't found anything strange or any bugs. Anything that stood out as weird to you?

vchuravy:
No I agreed with most of the changes. The only thing I was debating is that I tend to have the first arguments on the same line as the function.

```diff
- @kernel function lmem_copy_kernel!(output, @Const(input), 
-                                    ::Val{BANK}=Val(1)) where BANK
+ @kernel function lmem_copy_kernel!(
+         output, @Const(input),
+         ::Val{BANK} = Val(1),
+     ) where {BANK}
```

Now I probably should go in and use my delete key to write this as:

```
@kernel function lmem_copy_kernel!(
        output, @Const(input), ::Val{BANK} = Val(1),
    ) where {BANK}
```

fredrikekre:
Okay, yea lines are never joined automatically so that you can break lists like this the way you prefer.

The reason for not putting the first arguments on the first line is that the indent for the following line then become "function name dependent" at least if you want to line up the arguments.

PR #510, opened 2024-08-20
use github-action-benchmark

vchuravy:


avik-pal:
BTW [here](https://github.com/LuxDL/LuxLib.jl/pull/128/files) is a version of github-actions-benchmark that aggregates results by running the benchmarks on buidlkite. In my experience, running benchmarks on the github free runners gives very stochastic results.

PR #520, opened 2024-09-16
Remove test reliance on high-level features

maleadt:
This includes calls to `rand!`, `mapreducedim!`, type-mismatching constructors, etc. Even though most back-ends support these, KA.jl shouldn't assume so, making it easier to bootstrap a new platform.


maleadt:
Well, CI is a mess here... 😞 
All failures are consistent with what happens on master, so look unrelated to this PR.

PR #533, opened 2024-09-23
Start the long way to KernelAbstractions 1.0

vchuravy:
The kind of kernels that are expressible in KernelAbstractions are currently
limited by the CPU support having to opperate on a macro level.

With the CPU support being implemented in OpenCL.jl using POCL, we can drop
many restrictions.

### Breaking changes
- Removes CPU backend
- Removes `@uniform`
- Removes `@private`
- Julia LTS bump to 1.10

### TODOS
- [ ] POCL support on Windows
- [ ] Rework test
- [ ] Improve documentations
- [ ] Dynamic shared memory
- [ ] groupDim: `length(KernelAbstractions.blocks(KernelAbstractions.__iterspace(@context())))`
- [ ] Remove `return` limitation

### Open questions
- `CPU(;static=true)`
  - Interaction between POCL threads and Julia threads 
- How many macros do we really need?


src/KernelAbstractions.jl
```diff
@@ -465,12 +371,9 @@ function get_backend end
 # Should cover SubArray, ReshapedArray, ReinterpretArray, Hermitian, AbstractTriangular, etc.:
 get_backend(A::AbstractArray) = get_backend(parent(A))
 
-get_backend(::Array) = CPU()
-
```
vchuravy, line 469: Should this error?

jonas-schulze: If this fallback is deleted, what is the suggested alternative to `get_backend(A) == get_backend(B) || throw(ArgumentError(...))`?

Given #556, could this simply return `OpenCLBackend()` instead? Is it even necessary to distinguish between `CPU()` and `OpenCLBackend()`?

PR #534, opened 2024-09-29
Enzyme: fix propagation of runtime activity

wsmoses:
Required for chmy if differentiating an existing differentiated kernel


vchuravy:
We have to figure out the compat issue, currently there is no real reason for KA to stop supporting 1.6, besides the Enzyme extension

wsmoses:
I think the compat issue was introduced previously (presumably by me on the 0.13 enzyme PR). We should definitely fix it, but I presume this diff shouldn’t change things for it

PR #537, opened 2024-10-04
Enzyme support older versions

wsmoses:


ext/Enzyme012Ext.jl
```diff
@@ -0,0 +1,342 @@
+# https://github.com/EnzymeAD/Enzyme.jl/issues/1516
```
vchuravy, line 1: Can we name these files after the EnzymeCore version?

wsmoses: done

wsmoses:
is there a similar clang-format like command I can run to auto fix?

vchuravy:
Yeah you can https://github.com/fredrikekre/Runic.jl

```
julia --project=@runic -e 'using Pkg; Pkg.add(url = "https://github.com/fredrikekre/Runic.jl")'
alias runic="julia --project=@runic -e 'using Runic; exit(Runic.main(ARGS))' --"
 runic -i ext src
```

PR #538, opened 2024-10-18
Allow return statements for GPU-only kernels

pxl-th:
It's a bit annoying to not be able to use return statements.
Currently you have to do something like this, but it looks weird:

```julia
@kernel function ker!(args...)
    i = @index(Global)
    _ker!(i, args...)
end

function _ker!(i, args...)
    ...
end
```


vchuravy:
This needs tests. Making it GPU only is fine given #533

pxl-th:
@vchuravy if that's OK, I'd also like to tag a new release after this PR is merged.

vchuravy (approved):

PR #545, opened 2024-11-24
Add Atomix v1 compat

christiangnrd:
Not actually breaking since the only changes are the gpu backends from subpackages to extensions and it doesn't seem like AtomixCUDA was ever actually released.

Atomix v1 should be released once https://github.com/JuliaConcurrent/Atomix.jl/pull/43 is merged. (Edit: done)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,7 +20,7 @@ UnsafeAtomicsLLVM = "d80eeb9a-aca5-4d75-85e5-170c8b632249"
 
 [compat]
 Adapt = "0.4, 1.0, 2.0, 3.0, 4"
-Atomix = "0.1"
+Atomix = "0.1, 1"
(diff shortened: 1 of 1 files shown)


fhagemann:
`Atomix@1.0.0` introduced a `CUDA` extension, which results in a circular dependency `CUDA → KernelAbstractions → Atomix → CUDA`, see https://github.com/JuliaConcurrent/Atomix.jl/issues/47

christiangnrd:
> `Atomix@1.0.0` introduced a `CUDA` extension, which results in a circular dependency `CUDA → KernelAbstractions → Atomix → CUDA`, see [JuliaConcurrent/Atomix.jl#47](https://github.com/JuliaConcurrent/Atomix.jl/issues/47)

See https://github.com/JuliaConcurrent/Atomix.jl/issues/44

PR #550, opened 2024-12-14
attempt to force fastmath at the kernel level

leios:
I had a request to do #429 for fatmath, so here is my attempt.

2 issues:

1. My only issue is that I don't know what a good test should be, so I don't know if it works or not. Is there a good demo of some difference in precision from `@fastmath` and just normal execution?

2. I am also not sure what the equivalent lines are here (https://github.com/JuliaGPU/KernelAbstractions.jl/blob/main/src/macros.jl#L126):

```
    if force_inbounds
        push!(new_stmts, Expr(:inbounds, true))
    end
```

`Expr(:fastmath, true)` is not valid.

### DIFF
--- a/src/KernelAbstractions.jl
+++ b/src/KernelAbstractions.jl
@@ -50,7 +50,7 @@ synchronize(backend)
 ```
 """
 macro kernel(expr)
-    __kernel(expr, #=generate_cpu=# true, #=force_inbounds=# false)
+    __kernel(expr, #=generate_cpu=# true, #=force_inbounds=# false, #=force_fastmath=# false)
 end
 
 """
@@ -60,6 +60,7 @@ This allows for two different configurations:
 
 1. `cpu={true, false}`: Disables code-generation of the CPU function. This relaxes semantics such that KernelAbstractions primitives can be used in non-kernel functions.
 2. `inbounds={false, true}`: Enables a forced `@inbounds` macro around the function definition in the case the user is using too many `@inbounds` already in their kernel. Note that this can lead to incorrect results, crashes, etc and is fundamentally unsafe. Be careful!
+3. `fastmath={false, true}`: Enables a forced `@fastmath` macro around the function definition. This will use less precise square roots and flush denormals.
 
 - [`@context`](@ref)
 
@@ -72,6 +73,7 @@ macro kernel(ex...)
     else
         generate_cpu = true
         force_inbounds = false
+        force_fastmath = false
         for i in 1:(length(ex) - 1)
             if ex[i] isa Expr && ex[i].head == :(=) &&
                     ex[i].args[1] == :cpu && ex[i].args[2] isa Bool
@@ -79,16 +81,20 @@ macro kernel(ex...)
             elseif ex[i] isa Expr && ex[i].head == :(=) &&
                     ex[i].args[1] == :inbounds && ex[i].args[2] isa Bool
                 force_inbounds = ex[i].args[2]
+            elseif ex[i] isa Expr && ex[i].head == :(=) &&
+                    ex[i].args[1] == :fastmath && ex[i].args[2] isa Bool
+                force_fastmath = ex[i].args[2]
             else
                 error(
                     "Configuration should be of form:\n" *
                         "* `cpu=true`\n" *
                         "* `inbounds=false`\n" *
+                        "* `fastmath=false`\n" *
                         "got `", ex[i], "`",
                 )
             end
         end
-        __kernel(ex[end], generate_cpu, force_inbounds)
+        __kernel(ex[end], generate_cpu, force_inbounds, force_fastmath)
     end
 end
 
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -10,7 +10,7 @@ function find_return(stmt)
 end
 
 # XXX: Proper errors
-function __kernel(expr, generate_cpu = true, force_inbounds = false)
+function __kernel(expr, generate_cpu = true, force_inbounds = false, force_fastmath = false)
     def = splitdef(expr)
     name = def[:name]
     args = def[:args]
@@ -40,13 +40,13 @@ function __kernel(expr, generate_cpu = true, force_inbounds = false)
     if generate_cpu
         def_cpu = deepcopy(def)
         def_cpu[:name] = cpu_name
-        transform_cpu!(def_cpu, constargs, force_inbounds)
+        transform_cpu!(def_cpu, constargs, force_inbounds, force_fastmath)
         cpu_function = combinedef(def_cpu)
     end
 
     def_gpu = deepcopy(def)
     def_gpu[:name] = gpu_name = Symbol(:gpu_, name)
-    transform_gpu!(def_gpu, constargs, force_inbounds)
+    transform_gpu!(def_gpu, constargs, force_inbounds, force_fastmath)
     gpu_function = combinedef(def_gpu)
 
     # create constructor functions
@@ -78,7 +78,7 @@ end
 
 # The easy case, transform the function for GPU execution
 # - mark constant arguments by applying `constify`.
-function transform_gpu!(def, constargs, force_inbounds)
+function transform_gpu!(def, constargs, force_inbounds, force_fastmath)
     let_constargs = Expr[]
     for (i, arg) in enumerate(def[:args])
         if constargs[i]
@@ -90,6 +90,11 @@ function transform_gpu!(def, constargs, force_inbounds)
     if force_inbounds
         body = quote
             @inbounds $(body)
+        end
+    end
+    if force_fastmath
+        body = quote
+            @fastmath $(body)
         end
     end
     body = quote
@@ -112,7 +117,7 @@ end
 #   - handle indicies
 #   - hoist workgroup definitions
 #   - hoist uniform variables
-function transform_cpu!(def, constargs, force_inbounds)
+function transform_cpu!(def, constargs, force_inbounds, force_fastmath)
     let_constargs = Expr[]
     for (i, arg) in enumerate(def[:args])
         if constargs[i]
@@ -130,6 +135,7 @@ function transform_cpu!(def, constargs, force_inbounds)
     if force_inbounds
         push!(new_stmts, Expr(:inbounds, :pop))
     end
+
     push!(new_stmts, Expr(:popaliasscope))
     push!(new_stmts, :(return nothing))
     def[:body] = Expr(
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -22,7 +22,7 @@ end
 @test_throws ErrorException("This kernel is unavailable for backend CPU") my_no_cpu_kernel(CPU())
 
 # testing multiple configurations at the same time
-@kernel cpu = false inbounds = false function my_no_cpu_kernel2(a)
+@kernel cpu = false inbounds = false fastmath = false function my_no_cpu_kernel2(a)
 end
 @test_throws ErrorException("This kernel is unavailable for backend CPU") my_no_cpu_kernel2(CPU())
 
@@ -41,6 +41,16 @@ if Base.JLOptions().check_bounds == 0 || Base.JLOptions().check_bounds == 2
         a[idx] = 0
     end
     @test nothing == my_inbounds_kernel(CPU())(Int[], ndrange = 1)
+end
+
+if Base.JLOptions().fast_math == 0
+    @kernel fastmath = true function my_fastmath_kernel(a)
+        idx = @index(Global, Linear)
+        a[idx] = sqrt(10)
+    end
+    A = [0.0]
+    my_fastmath_kernel(CPU())(A, ndrange = 1)
+    @test A[1] == @fastmath sqrt(10)
 end
 
 struct NewBackend <: KernelAbstractions.GPU end


leios:
So for `transform_cpu(...)`, it should be something like...
```
    if force_fastmath
        push!(new_stmts, Expr(:macrocall, :@fastmath, arg2, arg3))
    end
```

But I don't know what `arg2`, and `arg3` are

I also don't know what to `:pop` below to mimic:

```
    if force_inbounds
        push!(new_stmts, Expr(:inbounds, :pop))
    end
```

I am working off of this:
```
julia> ex = :(@fastmath sqrt(1+1))
:(#= REPL[6]:1 =# @fastmath sqrt(1 + 1))

julia> ex.args
3-element Vector{Any}:
 Symbol("@fastmath")
 :(#= REPL[6]:1 =#)
 :(sqrt(1 + 1))

julia> ex.head
:macrocall
```

To figure out what the arguments should be

evelyne-ringoot:
Thank you! Here an example of a kernel that should return a different result when fastmath is turned on:
```
input=Float32.(ones(1e5))
input[1]=1e8

@kernel function nonassociative_sum!(input, output)
      sum =0.0f0;
      for i in 1:length(input)
            @inbounds sum+= input[i]
      end
      output[1]=sum
end
  
function my_nonassociative_sum!(input)
      backend = get_backend(input)
      output = KernelAbstractions.zeros(backend, Float32, 1)
  
      kernel = nonassociative_sum!(backend)
      kernel(input, output, ndrange = 1)
      return output
  end
 
  
my_nonassociative_sum!(input) #1e8
sum(input) #1e8+ 1e5
```

vchuravy:
The `fastmath` macro is sadly not region based like inbounds, but rather statement based...

```
@macroexpand @fastmath 1 + 1
```

```
:(Base.FastMath.add_fast(1, 1))
```

So you would need to wrap the entire code blocks inside the macro with `fastmath`. 

I, personally, have only made bad experiences with fastmath... 

In fact for @evelyne-ringoot example

```
@kernel function nonassociative_sum!(input, output)
      sum =0.0f0;
      for i in 1:length(input)
            @inbounds sum+= input[i]
      end
      output[1]=sum
end
```

I would prefer a:

```
@kernel function nonassociative_sum!(input, output)
      sum =0.0f0;
      @simd for i in 1:length(input)
            @inbounds sum+= input[i]
      end
      output[1]=sum
end
```

To mark the reduction chain as reassociatable.

leios:
tbh, I also have only had bad experiences with fastmath, so I am not sure if we should merge this in the end. I figured I would just get it out in the world and then we could discuss it further.

That said, it is an argument that is available from the `@cuda` launch macro that we do not have in KA. 

"So you would need to wrap the entire code blocks inside the macro with fastmath"
Like this on the GPU side?
```
   if force_fastmath
        body = quote
            @fastmath $(body)
        end
    end
```

If not, I guess we can:
1. dig into CUDA a bit to see how this is done, but I think it's on the NVIDIA side, not Julia side, so I don't think we can learn much from there
2. Ask users to just use `@fastmath` in front of specific lines they want fastmath for. I am a little afriad of something going awry somewhere else in the kernel because fastmath is enabled for the whole kernel.

vchuravy:
So `@cuda fastmath=true` is a weird beast. 

> - `fastmath`: use less precise square roots and flush denormals

It actually doesn't set fastmath on the IR level, but as a compiler flag see https://github.com/JuliaGPU/GPUCompiler.jl/pull/492 which then tells codegen to set flush-to-zero and prec_sqrt.

So that's quite a different beast than `@fastmath` in Julia. And indeed users can use these independently from each other, since cuda also supports `@fastmath`.

So this variant seems more like https://github.com/JuliaGPU/CUDA.jl/blob/4e9513b8a4e56629a236b58504d609b1775a8236/src/CUDAKernels.jl#L18

leios:
I guess there is no fastmath flag equivalent for metal / parallel cpu backends, so it is hard to set this in a generic way for everyone.

Maybe we should just ask @evelyne-ringoot (and other potential users) if it's fine to just `@fastmath` in front of specific lines instead of at the kernel level.

evelyne-ringoot:
From my side, I'm mostly looking to access the nvcc --fastmath equivalent in CUDA through KernelAbstractions, I am not sure whether this would work with the line-by-line solution?

vchuravy:
`nvcc --use_fast_math` is ill-defined. It is both a language level semantic change and a backend compiler change.

I think we can straight-forwardly expose the compiler semantics, but matching the language semantics is much harder.
Maybe we could gather examples with CUDA C in godbolt.org? And then look at what Julia + CUDA.jl create and if they match semantics.

evelyne-ringoot:
I've got an example where the julia compiler is not optimizing as much as nvcc and where Im not sure how to prompt it what to do differently. This is the [C++ version](https://godbolt.org/#z:OYLghAFBqRAWIDGB7AJgUwKKoJYBdkAnAGhxAgDMcAbdAOwEMBbdEAcgEY3iLk68Ayoga0QHACw8%2BeAKoBndAAUAHuwAM3AFZji1BnVCIArqgbEFtRHhx9y9VAGFk1I0zo77AGRx10AOVcAI3RCEAAmDmIAB2Q5fBs6Jxc3HRi46z5vXwCmYNCI83RLDLoBPAZCPCTXd0iLdCsEsoq8LP8gkPC68srqlO6Wtpy8roBKc2QjQkRWNgB6ACol5ZXVtfWNza2VuYBSNQBBRYBqRQY5PHRjvDgriiNqamOUfno8OWPkCmOAT0nCY5gNgAOmMQOOcn%2BM2OVFox1uhHQYDAxwWe0Oi22WOxOI26IOuzCAGYfIgXBhjrsiQ5EHBCHxkFTMPsCcTSeSrlSaRdcPwmSzCSS6GSTJzqYgeTZ%2BYdBezRZTxSYGAB9QhGfg4FjAuDS1lCkUUrkUC6I5i62XCjkKhw2E3oM1E5kytmW%2BVcpgMKLml0GsUOQj6VDIJje/VWrl4IxRWihuWG6lGazUfA/WOu%2BMOABuDQIhDTvutxkCejkyszYW1%2BfDish1GzhAAInRK46BT7qw45HaHU6Cc6iRgqL5rjR0HEAF5XIlhAUyw7K5XAajIQIiBfHTPIHCoY4LvTq2nKwKTOioUvKiBJsc4SejY6e6M/ACKeGVcQMtCvEGOczmFHOlwkPep7HOg2Z0DCAEhMcADu%2BBwJBFwejcLLHGhxw%2BHgEI3ugyo%2BMQqHoZh2GTsqkx4ARhzoRh/DHLgDDAHhgGUQc1HHs4nwUMa6B4PRwAsdRFDLgwWEvBcqLlEYAnoUJyAiai%2BEwsJWELORd67AA7AAQoRaHEcACoNscxbIIgADWACSqDKMCyhUjpVFEbROCGdcdL2qgVk2XZRIOaxMnKZ8iZyLsACsWlXhO6BhQ29m6bub5wBU6CoOusnyU%2BIXhZFOExXFjl6bRfEXC024%2BQ2fFMSECw5ZO%2BX%2BYVWFXl83F4FSDbQKMNWjlF9XUTg3wQC1Ci8TgDGjJpfnUdRzVcSNhI6USFVjcAJWVGVC21dFvnxZpsUFZSB3xbwAIQMR1DtWo9kXdSW3XQtC0Tdp8XUeRWVaRdoWxUt5FhVpEAXWEWnFT01jWV1UVkYmC04Jto7DTxeU7UdGn7Q1h19ujC5yD8wo3KaZ4QBNyOY4JRDHGdtFmZd9nU7dPU4QAtBwtMPUDT1TdNaGZX9sNfe1Ph/WdC0g6V4MLJDgtA9TQOi%2Bt1lI5zXPY7jVjuQwhPE0r03pU1TBRG%2BriuWowJqBQfVczC5OU1hN0NjLWks75N0OHdzts1pHMvZbeD64bIZA%2B1xw8%2BFn0NqpwV/WHFtc3t3tob7BtyK4D0/ZH4XU/zJOW9cfvJ0wCztZJQuVfgIQQzheF0AtmffdrZOned7UO07H0RgzdXu4H7OTfH01vVHMWM0XeeuAsIdt1n9foXHB2venWm18PS2J/7MfTSrePq5r6%2Bz6TMnW03S1Xe79O0L1Xc6T3z1z%2Bhv3hQDIsrWtYPKBXpG/UDsNA7NrVIw2A9Q6K12qjWcBJQFzkxm2Ac6AhxXDoK4fGHkPgzkgQuJcK41zKg3FuHcq4FDKgfNQZ8r53xLh4qOb8v5nAYABPWOIfBYLwQhNGfAjNqYHWIpLOg0lGokVwuRXhNEsKl2YvFdiTwEajQYkI3Wzw%2BDiRqgwKSx1AoLEUnIiOeB1I33RvpVyJlzJeVsuvYiLkg5II1sYny2tiKaFcpYzy1lgSpmzuhbGSVESpWwXIwBEUO7RX5tlAJwCDoeOSt4pScksITzukE/x59cpT3iuErxaVArCFpIEhJ146rxIQUwRxWU67xWInLV%2B7VRHVTdrY2iv95pLU6t1RJndbGDSkXxL2t8E7wzmojQOS1ykbR/iEtxaE97UTAZMg6J0KZHwbJoe6Z9cnbTbgMpsiDt5yC6ejee7xeZ5SAVnAB0MH7f2Bs/UGZV34COhkDQGOSpGhN2ehCefNYpHO%2Bg2QWZyn4MRftciWlcpZrIuf8q5CtkkozRtM/eaFN5qwJnIIm9VjqHypjTXydNXYBOZqzbunte7dKifJVe%2Bdjam3NmMnW6LbbNwWosoGrcXY1Iee1ApRSdk5xmqPAOi0jIT2jhnGKWj3pCq0u89ekyIEvLQpk24Bz4mLOOWSlO1LqIIqKSi9V6FVUhjTvs4V8TJU6rQrMm2xw7YnzbtSDlWz7oEq5dyhOvLU4NnldkyVnypUzxlTnPVhcV7KJLitKqhAblVxrs8nO5r5kO0ZY7JZOKWmrLZUtO1SKnXOr8eK95y8GwBsFYcxe0bLYTMtgNCmiylqXSzdyvxtdFUNnzXqn14y/WW01Vs7V2s95osbrRO21abXJpWQ6/lGbkF1r2ULc5QzxaQ0/qC%2BpiNjl%2BK9ZPEp0KwF7zYOMag7BQrcHcGwDQxBkDsAcDIBsrEHAewhFCTkxIuDEDwOoPdxB4BIGDFEUcJAyAQBQPrP9IA6CZkQIgDgYRxBGEiLCQCchyCBHfcQQIPgKg/HYC%2BtDjBCA/AAPKBE0DmLD3AgMsH4PhugxCUMYECEYYADgRDUEQ5wbgGAPQflmGe3AiJGjZlY2e9AygGiJlmC%2BzCRQUPJkCAGPDTgMAobwIQTUpHiD1mPAoBs6BONLh8KAU9H6hL/IAGo4HQDBfDUR6Bqd4PwIQIhWASCkPweQShVCGeINoSIegDBICVLoHAgREOQHGMgKIJRWOM3w0SY4jNGZGAIf%2BC4ypkIIXizydqiAPYcA0uheLiXcLJdfGl/LFRaTtTkEwZUAAOAAnNwZA9YVMYBC0TQoxQEh2FPH0WoxAvA%2BHaLkTokQ0jxD4L11IsRxt0CGB0UIdQig5iaKDSbi3Ot8GaJUObw2FvmFW84GoOgX47ZGBwcYkIpgzDEPuw9x6UMXrYMcPwJmHAOGOFB4E4hgQcAptgfA5NBTne4G%2Bwz4wv0YEzDgGYpByBAd/bQQgfhmCzBe29j7FZvsvsh9D1gigAAqAANfruBczbh0HZwQwhRDOcp25lQKHvMdeW7YCA9g1v9dPKdkb0RpslA52Nko3O9v1EaJtg7yQ%2Bui5KFt1og3hg85fhzk78v5s3dfYidA5O1C3bYEe4gJ6z2Pe01D6EZmLPQTR%2B9z7WO/uk8B8%2B4gxwnDAcR5SZ9owQfvtGOMMyYg1BqF0OwcQ93POPe4IhwPoONA%2B/UyEBh7hxBAA) , and the [julia version](https://github.com/ltorroba/cuda-qr/blob/multitile/julia_src/qr_kernels.jl#L137)  and [julia PTX](https://github.com/ltorroba/cuda-qr/blob/multitile/julia_src/julia_ptx).  Julia speed is approximately equal to C++ speed without fastmath, but fastmath increases speed by about a factor 1.5x.

I believe this has to do with the associative sums: in a previous version of the code, I had a split-k set-up, which improved performance in [julia](https://github.com/ltorroba/cuda-qr/blob/cd8b008368b9a0a4512fb7261bafce5d91a3fded/julia_src/qr_kernels.jl#L126) ([ptx](https://github.com/ltorroba/cuda-qr/blob/multitile/julia_src/julia_ptx#L314)), but did not make a difference in C++ (also on the goodbolt link).

Also I am interested in the fast sqrt for a different kernel in the same code! Generally nvcc --use_fast_math is an easily accesible flag (even if dangerous), while figuring out what parts of the code are optimized by it is less trivial. Perhaps a 'level' of fastmath could be a way to maintain code integrity while giving the user flexibility on how much accuracy/security they are willing to sacrifice for 'quick and dirty' performance?

evelyne-ringoot:
Hi, I took a look at the PTX, and it appears there is only some functions that are really different: 
-add.ftz.f32 vs add.f32
-fma instead of mul.f32 and add.f32
Both of which are covered by @cuda fastmath=true

The other differences can be covered by using the existing cpu=false  inbounds=true unsafe_indices=false flags, and using some of the other tips mentioned in https://cuda.juliagpu.org/stable/tutorials/performance/ ie 0-based indexing, while loops instead of ranges. 

@vchuravy Is it already possible to expose @cuda fastmath=true in KA in some way? In this case excluding the associative sum and other julia fastmath approximations as it appears in the PTX code I shared not to be happening (if I am reading it correctly).

PR #555, opened 2025-01-13
Document the semantics of copyto! and add pagelock!

vchuravy:


src/KernelAbstractions.jl
```diff
@@ -110,13 +110,45 @@ macro Const end
...
-Perform a `copyto!` operation that execution ordered with respect to the backend.
+Perform a `copyto!` operation that is execution ordered with respect to the backend.
+For most uses `Base.copyto!` provides a fully synchronous interface.
+
+!!! note
+    On some backends it may be necessary to first call [`pagelock!`](@ref) on host memory,
+    to enable fully asynchronous behaviour w.r.t to the host.
+
+!!! warning
+    If the function is asynchronous w.r.t to the host, the user is required to gurantuee, the lifetime
```
luraess, line 121:
```suggestion
    If the function is asynchronous w.r.t to the host, the user is required to guarantee, the lifetime
```

src/KernelAbstractions.jl
```diff
@@ -110,13 +110,45 @@ macro Const end
...
+    arr = zeros(64)
+    GC.@preserve arr begin
+        copyto!(backend, arr, ...)
+        # other operations
+        synchronize(backend)
+    end
+    ```
 
 !!! note
     Backend implementations **must** implement this function.
```
luraess, line 135:
```suggestion
    Backends **must** implement this function.
```
As to keep style with note for `pagelock!`

src/KernelAbstractions.jl
```diff
@@ -110,13 +110,45 @@ macro Const end
...
+    buffer can be freed.
+
+    ```julia
+    arr = zeros(64)
+    GC.@preserve arr begin
+        copyto!(backend, arr, ...)
+        # other operations
+        synchronize(backend)
+    end
+    ```
```
maleadt, line 132:
```suggestion
Perform an asynchronous `copyto!` operation that is execution ordered with respect to the back-end.

For most users, `Base.copyto!` should suffice, performance a simple, synchronous copy.
Only when you know you need asynchronicity w.r.t. the host, you should consider using
this asynchronous version, which requires additional lifetime guarantees as documented below.

!!! warning

    Because of the asynchronous nature of this operation, the user is required to guarantee that the lifetime
    of the source extends past the *completion* of the copy operation as to avoid a use-after-free. It is not
    sufficient to simply use `GC.@preserve` around the call to `copyto!`, because that only extends the
    lifetime past the operation getting queued. Instead, it may be required to `synchronize()`,
    or otherwise guarantee that the source will still be around when the copy is executed:

    ```julia
    arr = zeros(64)
    GC.@preserve arr begin
        copyto!(backend, arr, ...)
        # other operations
        synchronize(backend)
    end
    ```

!!! note

    On some back-ends it may be necessary to first call [`pagelock!`](@ref) on host memory
    to enable fully asynchronous behavior w.r.t to the host.
```

luraess: Further type-setting suggestions:

Perform an asynchronous `copyto!` operation that is **execution-ordered** with respect to the back-end.
For most users, `Base.copyto!` should suffice, **performing** a simple, synchronous copy.

luraess:
All tests currently fail because in line 139 we use `::Backend` which is only available after this point https://github.com/JuliaGPU/KernelAbstractions.jl/blob/d5c67f686497c5a2e44fcd64789080dff422d4f6/src/KernelAbstractions.jl#L452.

There were also changes and type-setting suggestions ☝️

PR #556, opened 2025-01-15
Implement a CPU backend using POCL

vchuravy:
TODO:

- [ ] POCL standalone build to remove dependency on `OpenCL_jll`
- [x] https://github.com/pocl/pocl/pull/1784 for support of exception handling


src/pocl/device/quirks.jl
```diff
@@ -0,0 +1,59 @@
+macro print_and_throw(args...)
+    return quote
+        @println "ERROR: " $(args...) "."
+        throw(nothing)
+    end
+end
+
+# math.jl
+@device_override @noinline Base.Math.throw_complex_domainerror(f::Symbol, x) =
```
vchuravy, line 9: @maleadt this will be fun.

PR #559, opened 2025-01-30
Implement groupreduce API

pxl-th:
Implement reduction API. Supports two types of algorithms:
- thread: reduction performed by threads: uses shmem of length `groupsize`, no bank conflict, no divergence.
- warp: reduction performed by `shlf_down` within warps: uses shmem of length `32`, reduction within warps storing results in shmem, followed by final warp reduction using values stored in shmem. Backends are required to only implement `shlf_down` intrinsic which AMDGPU/CUDA/Metal have (no sure about other backends).
- query function to check if backend supports warp reduction `KA.supports_warp_reduction(backend)`.

```julia
res = @groupreduce op val neutral Reduction.thread
res = @groupreduce op val neutral Reduction.warp
```

- Optionally limit number of threads that participate in reduction.
```julia
res = @groupreduce op val neutral Reduction.thread 128 # first 128 threads will perform reduction
```

### DIFF
--- a/src/reduce.jl
+++ b/src/reduce.jl
@@ -1,18 +1,28 @@
+export @groupreduce, Reduction
+
+module Reduction
+    const thread = Val(:thread)
+    const warp = Val(:warp)
+end
+
 """
-    @groupreduce algo op val neutral [groupsize]
+    @groupreduce op val neutral algo [groupsize]
 
 Perform group reduction of `val` using `op`.
 
 # Arguments
 
 - `algo` specifies which reduction algorithm to use:
-    - `:thread`:
+    - `Reduction.thread`:
         Perform thread group reduction (requires `groupsize * sizeof(T)` bytes of shared memory).
         Available accross all backends.
-    - `:warp`:
+    - `Reduction.warp`:
         Perform warp group reduction (requires `32 * sizeof(T)` bytes of shared memory).
+        Potentially faster, since requires fewer writes to shared memory.
+        To query if backend supports warp reduction, use `supports_warp_reduction(backend)`.
 
 - `neutral` should be a neutral w.r.t. `op`, such that `op(neutral, x) == x`.
+
 - `groupsize` specifies size of the workgroup.
     If a kernel does not specifies `groupsize` statically, then it is required to
     provide `groupsize`.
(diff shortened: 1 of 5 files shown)


pxl-th:
@vchuravy not sure about CPU errors. Any idea?

src/reduce.jl
```diff
@@ -0,0 +1,130 @@
...
+# Arguments
+
+- `algo` specifies which reduction algorithm to use:
+    - `Reduction.thread`:
+        Perform thread group reduction (requires `groupsize * sizeof(T)` bytes of shared memory).
+        Available accross all backends.
+    - `Reduction.warp`:
+        Perform warp group reduction (requires `32 * sizeof(T)` bytes of shared memory).
+        Potentially faster, since requires fewer writes to shared memory.
+        To query if backend supports warp reduction, use `supports_warp_reduction(backend)`.
```
vchuravy, line 22: Why is that needed? Shouldn't the backend go and use warp reductions if it can?

src/reduce.jl
```diff
@@ -0,0 +1,130 @@
...
+
+    s::UInt64 = groupsize ÷ 0x02
+    while s > 0x00
+        if (local_idx - 0x01) < s
+            other_idx = local_idx + s
+            if other_idx ≤ groupsize
+                @inbounds storage[local_idx] = op(storage[local_idx], storage[other_idx])
+            end
+        end
+        @synchronize()
```
vchuravy, line 77: Currently this is not legal.

#262 might need to wait until #556

vchuravy: (I assume this code is GPU only anyways)

test/groupreduce.jl
```diff
@@ -0,0 +1,51 @@
...
+    res = @groupreduce(op, val, neutral, algo)
+    i == 1 && (y[1] = res)
+end
+
+@kernel function groupreduce_2!(y, x, op, neutral, algo, ::Val{groupsize}) where {groupsize}
+    i = @index(Global)
+    val = i > length(x) ? neutral : x[i]
+    res = @groupreduce(op, val, neutral, algo, groupsize)
+    i == 1 && (y[1] = res)
+end
```
vchuravy, line 13: These need to be `cpu=false` since you are using non-top-level `@synchronize`

src/reduce.jl
```diff
@@ -0,0 +1,130 @@
...
+    return val
+end
+
+# Warp groupreduce.
+
+macro shfl_down(val, offset)
+    return quote
+        $__shfl_down($(esc(val)), $(esc(offset)))
+    end
+end
```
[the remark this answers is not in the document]
pxl-th: Yeah, removed macro.

test/groupreduce.jl
pxl-th: Done.

src/reduce.jl
pxl-th: I'm now doing an auto-selection of the algorithm based on device function `__supports_warp_reduction()`.

jonas-schulze:
I just saw that https://github.com/JuliaGPU/AMDGPU.jl/pull/729 has been closed. What is the prospect of this PR, or the general idea of a reduction abstraction to be added to KA.jl?

vchuravy:
> or the general idea of a reduction abstraction to be added to KA.jl?

That is still a goal of mine, but I am prioritizing getting the CPU backend to POCL done.

pxl-th:
https://github.com/JuliaGPU/AMDGPU.jl/pull/729 was closed, because from my testing I didn't see a major performance improvement from the warp reduce and in some cases (like [fused softmax](https://github.com/pxl-th/NNop.jl/blob/master/src/softmax.jl)) it was actually slower.
So I plan to just have shmem groupreduce API for now, unless people actually want warp-based reduction.

As for merging this PR my assumption was that #562 should go first, @vchuravy correct me if I'm wrong.

pxl-th:
> Apologies if my assessment is not entirely accurate as I am not intimately familiar with all the internal intricacies of `KernelAbstractions.jl`. I am implementing reductions for a different project and recalled that this draft PR exists. Disregarding the issues already raised with synchronisation primitives, I foresee that this implementation will face issues when `unsafe_indices = false` and the `ndrange` does not fit nicely into a multiple of whatever group size ends up being chosen, due to relying on uninitialised shared memory.
> 
> Imagine, for example, `ndrange = MAX_GROUP_SIZE + 4` for whatever the back-end's maximum group size is -- this launched two blocks of `1024` threads when testing on CUDA, with all but the first `4` threads of the second block being masked away, as confirmed via `@index(Local, Linear) == 1 && @print("group size = ", prod(@groupsize()))` and `@index(Group, Linear) == 2 && @print("index = ", @index(Local, Linear))`. In this case, the second block will perform a reduction on a `1024` item wide buffer in shared memory, but only the first `4` items will be properly initialised. The very first step of the reduction loop will have threads `1:4` combine their values with items `513:516`, which are just garbage memory. Being zero initialised by default does not help either, as there is no guarantee that the all-zero bit pattern will match the neutral element of the operation.

In this case `ndrange` should be a multiple of `groupsize` and you should perform boundschecking explicity.
See how I did this with online softmax [implementation](https://github.com/pxl-th/NNop.jl/blob/3df393d76506e861643114e58d6c2189fb306143/src/softmax.jl#L19).

PR #558, opened 2025-01-28
[0.10] Forbid divergent execution of work-group barriers

vchuravy:
As noted by @maleadt in https://github.com/JuliaGPU/OpenCL.jl/issues/283#issuecomment-2618785728

Several backends have a strict requirement that barriers like `@synchronize` must be executed convergent,
and the automatic bounds-checking in KA violates that.

Before GPU kernels lowered like:

```
if __validindex(__ctx__)
   # A
   @synchronize
   #B
end
```

Now they lower to:

```
__active_lane__ = __validindex(__ctx__)
if __active_lane__
   # A
end
@synchronize
if __active_lane__
   # B
end

PR #562, opened 2025-02-04
KernelIntrinsics

vchuravy:
The goal is to allow for kernels to be written without relying on KernelAbstractions macros

cc: @maleadt @pxl-th

### DIFF
--- a/src/intrinsics.jl
+++ b/src/intrinsics.jl
@@ -1,0 +1,52 @@
+module KernelIntrinsics
+
+"""
+    get_global_size()::@NamedTuple{x::Int32, y::Int32, z::Int32}
+
+Return the number of global work-items specified.
+
+!!! note
+    1-based.
+"""
+function get_global_size end
+
+"""
+    get_global_id()::@NamedTuple{x::Int32, y::Int32, z::Int32}
+
+Returns the unique global work-item ID.
+"""
+function get_global_id end
+
+"""
+    get_local_size()::@NamedTuple{x::Int32, y::Int32, z::Int32}
+
+Return the number of local work-items specified.
+"""
+function get_local_size end
+
+"""
+    get_local_id()::@NamedTuple{x::Int32, y::Int32, z::Int32}
+
+Returns the unique local work-item ID.
+"""
+function get_local_id end
+
+"""
+    get_num_groups()::@NamedTuple{x::Int32, y::Int32, z::Int32}
+
+Returns the number of groups.
+"""
+function get_num_groups end
+
+"""
+    get_group_id()::@NamedTuple{x::Int32, y::Int32, z::Int32}
+
+Returns the unique group ID.
+"""
+function get_group_id end
+
+function localmemory end
+function barrier end
+function print end
+
+end
--- a/src/KernelAbstractions.jl
+++ b/src/KernelAbstractions.jl
@@ -194,6 +194,10 @@ function unsafe_free! end
 
 unsafe_free!(::AbstractArray) = return
 
+include("intrinsics.jl")
+import .KernelIntrinsics
+export KernelIntrinsics
+
 ###
 # Kernel language
 # - @localmem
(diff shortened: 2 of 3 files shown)


src/intrinsics.jl
```diff
@@ -0,0 +1,52 @@
...
+Return the number of local work-items specified.
+"""
+function get_local_size end
+
+"""
+    get_local_id()::@NamedTuple{x::Int32, y::Int32, z::Int32}
+
+Returns the unique local work-item ID.
+"""
+function get_local_id end
```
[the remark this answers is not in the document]
vchuravy: Yeah basically, and my goal is to replace the old internal functions the people had to override with definitions based on these functions.

vchuravy:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/JuliaGPU/KernelAbstractions.jl/562?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#562** <a href="https://app.graphite.dev/github/pr/JuliaGPU/KernelAbstractions.jl/562?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/JuliaGPU/KernelAbstractions.jl/562?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#556** <a href="https://app.graphite.dev/github/pr/JuliaGPU/KernelAbstractions.jl/556?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

src/intrinsics.jl
```diff
@@ -0,0 +1,52 @@
...
+
+Return the number of global work-items specified.
+
+!!! note
+    1-based.
+"""
+function get_global_size end
+
+"""
+    get_global_id()::@NamedTuple{x::Int32, y::Int32, z::Int32}
```
vchuravy, line 14: Should this be Int32 or Int64?

maleadt:
So the idea is to decouple the back-ends from KA.jl, instead implementing KernelIntrinsics.jl? What's the advantage; do you envision packages other than KA.jl to build their kernel DSL on top of KernelIntrinsics.jl?

vchuravy:
> So the idea is to decouple the back-ends from KA.jl, instead implementing KernelIntrinsics.jl? What's the advantage; do you envision packages other than KA.jl to build their kernel DSL on top of KernelIntrinsics.jl?

I am very unsure if I am able to pull off #558 and as such, I want to remove the implicit `if validindex(ctx)` check,
and I want to make #559 possible.

My goal is to allow for a gracefull transition of the DSL macro language to something more similar to "just" OpenCL.
Removing the extra overhead introduced by the arbitrary dimensions, etc.

This would allow us to write performance critical kernels directly and solve the issue of how to write kernels that use barriers correctly. 

If I can figure out #558 the macro based DSL can stick around; otherwise I will encourage folks to move their kernel to `KernelIntrinsics`.

src/intrinsics.jl
vchuravy: OpenCL defines these as `Csize_t`

anicusan:
Will KA/KI still be a greatest common denominator of the GPU backends, or are you looking to introduce optional intrinsics? How will the groupreduce API do in terms of portability?

vchuravy:
> Will KA/KI still be a greatest common denominator of the GPU backends

The intrinsics proposed here are the greatest common denominator. I could see us adding some more intrinsics for reductions, but that is TBD.

src/intrinsics.jl
vchuravy: GPU dev call: We should make these `Int`/`size_t`

PR #583, opened 2025-03-12
Add Enzyme as a normal test dependency

vchuravy:


test/Project.toml
```diff
@@ -1,10 +1,11 @@
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
```
giordano, line 4: Haven't checked, but I presume `Enzyme` can be removed from https://github.com/JuliaGPU/KernelAbstractions.jl/blob/292269fafef68bd2610316af55acfddb178c714e/test/quality_assurance.jl#L7-L8

vchuravy:
x-ref: https://github.com/EnzymeAD/Enzyme.jl/pull/3309

PR #593, opened 2025-05-02
[release-0.9] Test correct backend with examples tests

christiangnrd:
Made against `release-0.9` so tests will actually run.

Before:
```
From worker 2:	(backend, backend_str) = (KernelAbstractions.CPU(false), "oneAPI") # oneAPI
From worker 2:	(backend, backend_str) = (KernelAbstractions.CPU(false), "Metal") # Metal
From worker 2:	(backend, backend_str) = (CUDABackend(false, false), "CUDA") # CUDA
Succeeds? # AMD
```

After:
```
From worker 2:	(backend, backend_str) = (oneAPIBackend(), "oneAPI") # oneAPI
From worker 2:	(backend, backend_str) = (MetalBackend(), "Metal") # Metal
From worker 2:	(backend, backend_str) = (CUDABackend(false, false), "CUDA") # CUDA
Succeeds? # AMD
```

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -11,6 +11,7 @@ steps:
       - JuliaCI/julia-coverage#v1:
           codecov: true
     command: |
+      echo -e "[CUDA_Runtime_jll]\nlocal = \"true\"" >LocalPreferences.toml
       julia -e 'println("--- :julia: Instantiating project")
                 using Pkg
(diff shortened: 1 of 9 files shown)


christiangnrd:
Ok Seems like CUDA is still broken, I'll wait for that to be fixed. I don't understand the oneAPI failure in 1.11. Maybe it would be fixed on the main branch? However I can't test that easily.

PR #599, opened 2025-05-20
Switch to pocl_jll@v7

vchuravy:
- **use pocl_jll @v7**
- **Set CI**


christiangnrd:
You should be able to get the ubuntu arm tests passing by registering the already bumped SPIRVIntrinsics.jl

PR #608, opened 2025-06-10
Enable downstream CI for 0.10

vchuravy:


vchuravy:
@maleadt Currently we only run the KA tests for CUDA, but should we also run some GPUArrays test to ensure that KA doesn't break things?

maleadt:
> should we also run some GPUArrays test to ensure that KA doesn't break things?

That would probably be smart, yeah. Maybe once the dust settles and JLArrays can use the new CPU back-end?

PR #622, opened 2025-06-30
[0.10] add get_backend for StaticArrays

vchuravy:
Port of #621


Project.toml
```diff
@@ -50,6 +51,7 @@ julia = "1.10"
 pocl_jll = "7"
 
 [extras]
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 EnzymeCore = "f151be2c-9106-41f4-ab19-57ee4f262869"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
```
christiangnrd, line 57:
```suggestion
```
Since test/Porject.toml exists, I think this can be deleted and the `StaticArrays` entry added to that file instead

test/test.jl
```diff
@@ -95,6 +96,18 @@ function unittest_testsuite(Backend, backend_str, backend_mod, BackendArrayT; sk
         @test @inferred(KernelAbstractions.get_backend(sparse(A))) isa backendT
     end
 
+    @conditional_testset "StaticArrays" skip_tests begin
+        backend = Backend()
+        backendT = typeof(backend).name.wrapper # To look through CUDABackend{true, false}
+        @test backend isa backendT
+
+        @test KernelAbstractions.get_backend(@MMatrix [1.0]) isa CPU
```
christiangnrd:
```suggestion
        @test KernelAbstractions.get_backend(@MMatrix [1.0]) isa KernelAbstractions.POCLBackend
```

christiangnrd:
Guess we need to find an alternative to `MArray` for `KA.scratchpad` or re-add StaticArrays.jl as a full dependency.
https://github.com/JuliaGPU/KernelAbstractions.jl/blob/44097082d2b3e5931bcbb9c3275e58bc5f0e10ca/src/pocl/backend.jl#L224-L226

PR #625, opened 2025-07-04
avoid boxing when `@kernel` is used as a closure

simeonschaub:
This allows uses of `@kernel` inside of functions without running into
JuliaLang/julia#53295. One thing I have been wondering though: Is the
`@isdefined` check here really necessary? In global scope it should just
override the previous definition and for closures these kind of if
statements are ignored anyways.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -50,12 +50,10 @@ function __kernel(expr, generate_cpu = true, force_inbounds = false, unsafe_indi
     gpu_function = combinedef(def_gpu)
 
     # create constructor functions
+    _name = Symbol(:_, name)
     constructors = quote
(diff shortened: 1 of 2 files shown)


vchuravy:
This was added in https://github.com/JuliaGPU/KernelAbstractions.jl/commit/2896ef745217ac37cc53be1dd296ca7f264e5f25

Which probably was meant to avoid re-definition warnings when defining a kernel with multiple methods.

vchuravy:
Should be okay. @maleadt has asked for this before.

vchuravy:
Can you also add a PR against main?

simeonschaub:
Will do!

PR #632, opened 2025-07-21
[0.9] Unified memory allocations

christiangnrd:
See #630 

Feedback from #630 has been addressed. When not implemented in the backend, no kwarg or `unified=false` just work, while `unified=true` will fail unless implemented.

### DIFF
--- a/src/KernelAbstractions.jl
+++ b/src/KernelAbstractions.jl
@@ -587,8 +587,7 @@ Returns whether unified memory arrays are supported by the backend.
 
 !!! note
     Backend implementations **should** implement this function
-    only if they **do** support unified memory. It will be required
-    in KernelAbstractions 0.10.
+    only if they **do** support unified memory.
 """
 supports_unified(::Backend) = false
 
--- a/src/KernelAbstractions.jl
+++ b/src/KernelAbstractions.jl
@@ -534,9 +534,9 @@ Adapt.adapt_storage(::CPU, a::Array) = a
 """
     allocate(::Backend, Type, dims...; unified=false)::AbstractArray
 
-Allocate a storage array appropriate for the computational backend. `unified`
-allocates an array using unified memory if the backend supports it. Use
-[`supports_unified`](@ref) to determine whether it is supported by a backend.
+Allocate a storage array appropriate for the computational backend. `unified=true`
+allocates an array using unified memory if the backend supports it and throws otherwise.
+Use [`supports_unified`](@ref) to determine whether it is supported by a backend.
 
 !!! note
     Backend implementations **must** implement `allocate(::NewBackend, T, dims::Tuple)`
@@ -558,7 +558,8 @@ end
     zeros(::Backend, Type, dims...; unified=false)::AbstractArray
 
 Allocate a storage array appropriate for the computational backend filled with zeros.
-`unified` allocates an array using unified memory if the backend supports it.
+`unified=true` allocates an array using unified memory if the backend supports it and
+throws otherwise.
 """
 zeros(backend::Backend, T::Type, dims...; kwargs...) = zeros(backend, T, dims; kwargs...)
 function zeros(backend::Backend, ::Type{T}, dims::Tuple; kwargs...) where {T}
@@ -571,7 +572,8 @@ end
     ones(::Backend, Type, dims...; unified=false)::AbstractArray
 
 Allocate a storage array appropriate for the computational backend filled with ones.
-`unified` allocates an array using unified memory if the backend supports it.
+`unified=true` allocates an array using unified memory if the backend supports it and
+throws otherwise.
 """
 ones(backend::Backend, T::Type, dims...; kwargs...) = ones(backend, T, dims; kwargs...)
 function ones(backend::Backend, ::Type{T}, dims::Tuple; kwargs...) where {T}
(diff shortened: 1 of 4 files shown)


christiangnrd:
Errors unrelated?

src/KernelAbstractions.jl
```diff
@@ -532,40 +532,64 @@ get_backend(::Array) = CPU()
...
+[`supports_unified`](@ref) to determine whether it is supported by a backend.
 
 !!! note
     Backend implementations **must** implement `allocate(::NewBackend, T, dims::Tuple)`
 """
-allocate(backend::Backend, T::Type, dims...) = allocate(backend, T, dims)
-allocate(backend::Backend, T::Type, dims::Tuple) = throw(MethodError(allocate, (backend, T, dims)))
+allocate(backend::Backend, T::Type, dims...; kwargs...) = allocate(backend, T, dims; kwargs...)
+function allocate(backend::Backend, T::Type, dims::Tuple; unified::Union{Nothing, Bool} = nothing)
+    return if isnothing(unified)
```
vchuravy, line 546: Please move the return into the if

src/KernelAbstractions.jl
```diff
@@ -532,40 +532,64 @@ get_backend(::Array) = CPU()
...
+        throw(MethodError(allocate, (backend, T, dims)))
+    elseif unified
+        throw(ArgumentError("`$(typeof(backend))` either does not support unified memory or it has not yet defined `allocate(backend::$backend, T::Type, dims::Tuple; unified::Bool)`"))
```
vchuravy, line 549: We can use the query to check?

src/KernelAbstractions.jl
```diff
@@ -532,40 +532,64 @@ get_backend(::Array) = CPU()
...
+allocates an array using unified memory if the backend supports it. Use
+[`supports_unified`](@ref) to determine whether it is supported by a backend.
 
 !!! note
     Backend implementations **must** implement `allocate(::NewBackend, T, dims::Tuple)`
 """
-allocate(backend::Backend, T::Type, dims...) = allocate(backend, T, dims)
-allocate(backend::Backend, T::Type, dims::Tuple) = throw(MethodError(allocate, (backend, T, dims)))
+allocate(backend::Backend, T::Type, dims...; kwargs...) = allocate(backend, T, dims; kwargs...)
+function allocate(backend::Backend, T::Type, dims::Tuple; unified::Union{Nothing, Bool} = nothing)
```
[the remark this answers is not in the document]
christiangnrd: It defaults to `nothing` so that if the no-kwargs method is called for a backend that doesn't have `allocate` defined, it throws, and if you call `allocate` with `unified=false` on a backend that only has the no-kwarg version defined, it works.

src/KernelAbstractions.jl
christiangnrd: I had the following:
```suggestion
        if supports_unified(backend)
            throw(ArgumentError("`$(typeof(backend))` states support for unified memory, but has not defined `allocate(backend::$(typeof(backend)), T::Type, dims::Tuple; unified::Bool)`. Pleas "))
        else
            throw(ArgumentError("`$(typeof(backend))` does not support unified memory. If you believe it does, please open a github issue requesting adding support for it."))
        end
```

but this feels really complicated for no reason. I think 

```suggestion
        throw(ArgumentError("`$(typeof(backend))` does not support unified memory. If you believe it should, please open a github issue requesting adding support for it."))
```
would work since whether a device supports it or not, it's true that the julia KA backend for it does not in the situation where a user is encountering this error.

src/KernelAbstractions.jl
```diff
@@ -532,40 +532,64 @@ get_backend(::Array) = CPU()
...
-    allocate(::Backend, Type, dims...)::AbstractArray
+    allocate(::Backend, Type, dims...; unified=false)::AbstractArray
 
-Allocate a storage array appropriate for the computational backend.
+Allocate a storage array appropriate for the computational backend. `unified`
+allocates an array using unified memory if the backend supports it. Use
+[`supports_unified`](@ref) to determine whether it is supported by a backend.
 
 !!! note
     Backend implementations **must** implement `allocate(::NewBackend, T, dims::Tuple)`
```
vchuravy, line 542:
```suggestion
    Backend implementations **must** implement `allocate(::NewBackend, T, dims::Tuple)`
    Backend implementations **should** implement `allocate(::NewBackend, T, dims::Tuple; unified::Bool=false)`
```

src/KernelAbstractions.jl
```diff
@@ -532,40 +532,66 @@ get_backend(::Array) = CPU()
...
 
+"""
+    supports_unified(::Backend)::Bool
+
+Returns whether unified memory arrays are supported by the backend.
+
+!!! note
+    Backend implementations **should** implement this function
+    only if they **do** support unified memory. It will be required
+    in KernelAbstractions 0.10.
```
vchuravy, line 591:
```suggestion
    only if they **do** support unified memory.
```

I would rather not state anything as of yet in the docs about 0.10

src/KernelAbstractions.jl
```diff
@@ -532,40 +532,65 @@ get_backend(::Array) = CPU()
...
+    end
+end
+
 
 """
-    zeros(::Backend, Type, dims...)::AbstractArray
+    zeros(::Backend, Type, dims...; unified=false)::AbstractArray
 
 Allocate a storage array appropriate for the computational backend filled with zeros.
+`unified` allocates an array using unified memory if the backend supports it.
```
christiangnrd, line 561: @vchuravy Should these mention that `unified=true` on an unsupported backed throws an error?

christiangnrd: Done!

vchuravy:
The docs error is introduced. You need to add `supports_unified` to the API docs.

PR #635, opened 2025-09-30
More KernelIntrinsics

christiangnrd:
Ideally would target a rebased vc/intrinsics branch, but targets main for now.

@vchuravy @maleadt


vchuravy:
Can you rebase?

src/intrinsics.jl
```diff
@@ -0,0 +1,357 @@
...
+
+!!! note
+    Backend implementations **must** implement:
+    ```
+    @device_override barrier()
+    ```
+"""
+function barrier()
+    error("Group barrier used outside kernel or not captured")
+end
```
vchuravy: We should specify the memory semantics that are required by the barrier implementation

src/intrinsics.jl
```diff
@@ -0,0 +1,357 @@
...
+    ```
+    As well as the on-device functionality.
+"""
+struct KIKernel{B, Kern}
+    backend::B
+    kern::Kern
+end
+
+"""
+    kernel_max_work_group_size(backend, kern; [max_work_items::Int])::Int
```
vchuravy: Do we need `backend` as a first argument?

src/intrinsics.jl
```diff
@@ -0,0 +1,357 @@
...
+This function is called for every argument to be passed to a kernel, allowing it to be
+converted to a GPU-friendly format.
+
+!!! note
+    Backend implementations **must** implement:
+    ```
+    kiconvert(::NewBackend, arg)
+    ```
+"""
+function kiconvert end
```
vchuravy: Naming nit again. Maybe call it `argconvert`

src/KernelAbstractions.jl
```diff
@@ -239,11 +248,8 @@ end
 Declare storage that is local to a workgroup.
 """
 macro localmem(T, dims)
-    # Stay in sync with CUDAnative
-    id = gensym("static_shmem")
-
     return quote
-        $SharedMemory($(esc(T)), Val($(esc(dims))), Val($(QuoteNode(id))))
+        $SharedMemory($(esc(T)), Val($(esc(dims))))
```
vchuravy: This is technically an ABI break, which I had avoided so far.

src/intrinsics.jl
christiangnrd: Nope!

src/intrinsics.jl
```diff
@@ -0,0 +1,357 @@
...
+        if T <: Val
+            item = QuoteNode(T.parameters[1])
+        end
+        push!(args, item)
+    end
+
+    return quote
+        print($(args...))
+    end
+end
```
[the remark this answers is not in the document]
christiangnrd: I just took this from the KA implementation. I assume this was the old cpu backend code? If so I'll remove

src/intrinsics.jl
vchuravy: The semantics in the docs string should be more specific

src/intrinsics.jl
```diff
@@ -0,0 +1,342 @@
...
+    ```
+    If the backend does not support printing,
+    define it to return `nothing`.
+"""
+function _print end
+
+
+
+"""
+    Kernel{Backend, BKern}
```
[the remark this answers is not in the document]
christiangnrd: Old placeholder name that stood for backend kernel. Good catch

src/intrinsics.jl
```diff
@@ -0,0 +1,342 @@
+module KernelIntrinsics
```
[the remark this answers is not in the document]
christiangnrd: I've thrown up something. Very open to feeedback

christiangnrd:
I'm thinking we should either directly rename `KernelIntrinsics` to `KI` or provide the alias.

src/intrinsics.jl
vchuravy: In particular:
- https://github.com/JuliaGPU/KernelAbstractions.jl/pull/618
- https://github.com/JuliaGPU/OpenCL.jl/pull/338
- https://github.com/JuliaGPU/AMDGPU.jl/pull/783
- https://github.com/JuliaGPU/Metal.jl/issues/608

src/intrinsics.jl
```diff
@@ -0,0 +1,363 @@
...
+Keyword arguments:
+- `name`: override the name that the kernel will have in the generated code
+
+!!! note
+    Backend implementations **must** implement:
+    ```
+    gpufunction(::NewBackend, f::F, tt::TT=Tuple{}; name=nothing, kwargs...) where {F,TT}
+    ```
+"""
+function gpufunction end
```
vchuravy: Should we just call this `kernel_function`?

christiangnrd:
> Do we really need to, given the `import KernelIntrinsics as KI` feature?

I felt like we might as well include it since it'll be written so much, and I found fully qualifying `KernelIntrinsics` for every KI function mention in the docstrings was distracting, but also didn't want to write `KI` since users new to the package would be confused wondering where it came from.

I don't feel that strongly about this though so I'll defer to you for the final decision

maleadt:
Yeah, I get that argument, but it still seems like a very generic name to export. I'll defer to @vchuravy on this :-)

PR #648, opened 2025-10-24
Update quickstart.md to include MetalBackend (docs)

vcantarella:
Added syntax for using the Metal backend in kernel launching section (docs).

### DIFF
--- a/docs/src/quickstart.md
+++ b/docs/src/quickstart.md
@@ -74,7 +74,7 @@ The kernel generation and execution are then
 backend = get_backend(A)
 mul2_kernel(backend, 64)(A, ndrange=size(A))
 synchronize(backend)
-all(A .== 2.0) # For Metal use: all(A .== 2.0f0) : only float32 are supported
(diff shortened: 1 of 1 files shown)


docs/src/quickstart.md
```diff
@@ -64,12 +64,17 @@ A = ROCArray(ones(1024, 1024))
...
+using Metal: MtlArray
+A = MtlArray(Float32.(ones(1024, 1024)))
+```
 The kernel generation and execution are then
 ```julia
 backend = get_backend(A)
 mul2_kernel(backend, 64)(A, ndrange=size(A))
 synchronize(backend)
-all(A .== 2.0)
+all(A .== 2.0) # For Metal use: all(A .== 2.0f0) : only float32 are supported
```
christiangnrd: This way we don't need to add a special case comment.
```suggestion
all(A .≈ 2)
```

christiangnrd:
Thanks!

PR #659, opened 2025-11-19
[pocl] RNG support

### LINKED ISSUE #641: [pocl] Device side RNG
We currently don't have a test for this, but as soon as all of the backends support device side rng we should make that a requirement.

For POCL we likely have to copy https://github.com/JuliaGPU/OpenCL.jl/pull/365

simeonschaub:
closes #641

Testing locally, I am running into #624

### DIFF
--- a/src/pocl/device/random.jl
+++ b/src/pocl/device/random.jl
@@ -1,0 +1,234 @@
+## random number generation
+
+using Random
+import RandomNumbers
+
+# local memory with the actual seed, per subgroup, set by `initialize_rng_state`` or overridden by calling `seed!`
+@inline function global_random_keys()
+    n = get_num_sub_groups()
+    ptr = random_keys()::LLVMPtr{UInt32, AS.Workgroup}
+    return CLDeviceArray{UInt32, 1, AS.Workgroup}((n,), ptr)
+end
+
+# local memory with per-subgroup counters, incremented when generating numbers
+@inline function global_random_counters()
+    n = get_num_sub_groups()
+    ptr = random_counters()::LLVMPtr{UInt32, AS.Workgroup}
+    return CLDeviceArray{UInt32, 1, AS.Workgroup}((n,), ptr)
+end
+
+# initialization function, called automatically at the start of each kernel
+function initialize_rng_state()
+    subgroup_id = get_sub_group_id()
+    @inbounds global_random_keys()[subgroup_id] = kernel_state().random_seed
+    @inbounds global_random_counters()[subgroup_id] = 0
+end
+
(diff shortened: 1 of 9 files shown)


test/random.jl
```diff
@@ -0,0 +1,166 @@
...
+        # should result in the same numbers
+        Random.seed!(seed)
+    elseif seed === nothing
+        # should result in different numbers across launches,
+        # as determined by the seed set during module loading.
+    end
+end
+
+function random_testsuite(backend)
+    eltypes = [Float16, Float32, Float64, Int32, UInt32, Int64, UInt64, Bool, UInt16]
```
christiangnrd:
```suggestion
    eltypes = [Float16, Float32, Int32, UInt32, Int64, UInt64, Bool, UInt16]
    KA.supports_float64(backend()) && push!(eltypes, Float64)
```

christiangnrd: Do we also need a `supports_float16`?

vchuravy:
Can you rebase this?

test/random.jl
simeonschaub: Yes, I think that would be sensible. I have commented out Float16 for now, since I don't think our pocl build supports it

simeonschaub:
Done

src/pocl/device/random.jl
```diff
@@ -0,0 +1,267 @@
...
+    # ... where it used to call `Random_make_seed()`
+    @device_override Random.make_seed() = kernel_state().random_seed
+end
+
+"""
+    Random.rand(rng::Philox2x32, UInt32)
+
+Generate a byte of random data using the on-device Tausworthe generator.
+"""
+function Random.rand(rng::Philox2x32{R}, ::Type{UInt64}) where {R}
```
vchuravy, line 98: Isn't this type pirarcy?

simeonschaub: The type `Philox2x32` is defined above, so it shouldn't be?

vchuravy: I was wondering why we had to reimplement some of the random number generator ourselves instead of just being able to use the library, can you add some comments on why that is necessary, and what the key differences are?

simeonschaub: This part is mostly just copied from CUDA.jl, so I don't know the original justification exactly. AFAIU, the implementation in Random123 is not GPU-compatible, so that's why we need to reimplement some of it ourselves. I think it would be a good idea to eventually have a library for this, which all of the GPU packages can then depend on to avoid some of the duplication, but for now this seems like the easiest solution

PR #663, opened 2025-11-28
Update CI workflow to use Windows 2022

christiangnrd:
See https://github.com/JuliaGPU/OpenCL.jl/issues/393 and https://github.com/JuliaGPU/OpenCL.jl/pull/405


christiangnrd:
AMDGPU failures expected and unrelated

Issue #665, opened 2025-12-10
Unroll factor compilation not available for Val values


evelyne-ringoot:
I am trying to do the following:
```
using KernelAbstractions, CUDA
using KernelAbstractions.Extras: @unroll
backend = CUDABackend()

@kernel function add1_kernel!(A, ::Val{VALSPERTHREAD} = Val(4), ::Val{UNROLLFACTOR} = Val(4)) where {VALSPERTHREAD, UNROLLFACTOR}
    I = @index(Global,Linear)
    @unroll UNROLLFACTOR for i in 0:VALSPERTHREAD-1
        @inbounds A[I*4+i] = A[I*4+i]+1
    end
end

function myadd!(A)
    backend = get_backend(A)
    kernel = add1_kernel!(backend)
    kernel(A, ndrange = (Int(length(A)/4)))
    return
end

A = KernelAbstractions.zeros(backend, Float32, 64)
myadd!(A)
KernelAbstractions.synchronize(backend)
```

And getting an error:
```
ERROR: LoadError: Syntax error: `@unroll N expr` needs a constant integer N
Stacktrace:
 [1] error(s::String)
   @ Base .\error.jl:35
 [2] var"@unroll"(__source__::LineNumberNode, __module__::Module, N::Any, expr::Any)
   @ KernelAbstractions.Extras.LoopInfo C:\Users\evely\.julia\packages\KernelAbstractions\lGrz7\src\extras\loopinfo.jl:60
in expression starting at C:\Users\evely\NextLATuner\juliakernels\unrollparam.jl:7
in expression starting at C:\Users\evely\NextLATuner\juliakernels\unrollparam.jl:5
```

PR #667, opened 2025-12-15
support emitting a generated function

vchuravy:
Motivated by #665

Proposed syntax

```
@kernel generated=true function f(::Val{N}) where N
           KernelAbstractions.Extras.@unroll $N for i in 1:10
           end
       end
```

Sadly this doesn't quite work yet, since I need to handle the `$N` correctly

Currently:

```
    function cpu_f(__ctx__, ::Val{N}; ) where N
        if $(Expr(:generated))
            $(Expr(:copyast, :($(QuoteNode(:(let
      begin
          var"##N#249" = length((KernelAbstractions.__workitems_iterspace)(__ctx__))
          begin
              #= /home/vchuravy/src/KernelAbstractions/src/macros.jl:317 =#
              for var"##I#248" = (KernelAbstractions.__workitems_iterspace)(__ctx__)
                  #= /home/vchuravy/src/KernelAbstractions/src/macros.jl:318 =#
                  (KernelAbstractions.__validindex)(__ctx__, var"##I#248") || continue
                  #= /home/vchuravy/src/KernelAbstractions/src/macros.jl:319 =#
                  #= /home/vchuravy/src/KernelAbstractions/src/macros.jl:320 =#
                  #= REPL[19]:2 =# KernelAbstractions.Extra.@unroll $(Expr(:$, :N)) for i = 1:10
                          #= REPL[19]:2 =#
                          #= REPL[19]:3 =#
                      end
                  #= /home/vchuravy/src/KernelAbstractions/src/macros.jl:321 =#
              end
          end
      end
      return nothing
  end))))))
        else
            $(Expr(:meta, :generated_only))
        end
    end
```

Whereas

```
julia> @macroexpand @generated function(::Val{N}) where N
           quote
               KernelAbstractions.Extras.@unroll $N for i in 1:10
               end
           end
       end
:(function (::Val{N},) where N
      #= REPL[18]:1 =#
      if $(Expr(:generated))
          #= REPL[18]:1 =#
          #= REPL[18]:2 =#
          Core._expr(:block, $(QuoteNode(:(#= REPL[18]:3 =#))), Core._expr(:macrocall, $(Expr(:copyast, :($(QuoteNode(:(KernelAbstractions.Extras.var"@unroll")))))), $(QuoteNode(:(#= REPL[18]:3 =#))), N, $(Expr(:copyast, :($(QuoteNode(:(for i = 1:10
      #= REPL[18]:3 =#
      #= REPL[18]:4 =#
  end))))))))
      else
          $(Expr(:meta, :generated_only))
          return
      end
  end)
```

Note that the QuoteNode got broken into smaller pieces with the interpolated variable being left alone and the rest being passed to `_expr` and `QuoteNode`

### DIFF
--- a/src/KernelAbstractions.jl
+++ b/src/KernelAbstractions.jl
@@ -50,7 +50,7 @@ synchronize(backend)
 ```
 """
 macro kernel(expr)
-    return __kernel(expr, #=generate_cpu=# true, #=force_inbounds=# false, #=unsafe_indices=# false)
+    return __kernel(expr, #=generate_cpu=# true, #=force_inbounds=# false, #=unsafe_indices=# false, #=generated=# false)
 end
 
 """
@@ -69,11 +69,12 @@ This allows for two different configurations:
 """
 macro kernel(ex...)
     if length(ex) == 1
-        return __kernel(ex[1], true, false, false)
+        return __kernel(ex[1], true, false, false, false)
     else
         generate_cpu = true
         unsafe_indices = false
         force_inbounds = false
+        generated = false
         for i in 1:(length(ex) - 1)
             if ex[i] isa Expr && ex[i].head == :(=) &&
                     ex[i].args[1] == :cpu && ex[i].args[2] isa Bool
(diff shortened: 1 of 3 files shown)


fjwillemsen:
@vchuravy can this be merged? We would like to publish the package soon 😄

vchuravy:
@fjwillemsen nobody expressed to me after my proposal that they actually wanted this, and I don't like to merge functionality without at least a clear user.

fjwillemsen:
I understand, I'll get @evelyne-ringoot in the loop and get back to you on this. Thank you!

evelyne-ringoot:
Hi, the original motivation for this was the performance difference in all kernels where input values are being used for loop unrolls, versus where this values is defined as const, so the scope is not limited to specifying the loop unroll number, but general availability of loop unrolling constants during early compile times, in particular for more complex functions. However, I can no longer reproduce a substantial performance difference in all kernels but one, where the performance difference swings both ways depending on hyperparameters (better using ::Val vs better using const). Considering that, the master branch works fine for our application!

PR #668, opened 2025-12-23
KI Sub-groups

christiangnrd:


test/intrinsics.jl
```diff
@@ -1,21 +1,61 @@
...
+end
+
+function shfl_down_test_kernel(a, b, ::Val{N}) where N
+    idx = KI.get_sub_group_local_id()
+
+    val = a[idx]
+
+    offset = 0x00000001
+    while offset < N
+        val += KI.shfl_down(val, offset)
```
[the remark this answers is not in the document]
christiangnrd: Seems like all sub-group implementations are inherently synchronizing by default (except for CUDA but the non-synchronizing version is deprecated) so that's what we're going with.

christiangnrd:
This is mostly ready except that for correctness of algorithms, I think it requires us to know the subgroup size of a device at compile time. I tried to force a known subgroup size for openCL/pocl, but that seems to have broken on-kernel rand. I'm not sure how to best approach this. I currently work around this in OpenCL.jl to test that the rest works by forcing a workgroup size in the KI launch interface so that the non-KI rand kernel doesn't encounter this issue, but that's a hack.

I think the ideal solution would be to have `get_num_sub_groups` always return at least 1 so we can force a sub-group size.

christiangnrd:
@maleadt @vchuravy This is ready for final review. If/once accepted I’ll merge the backend subgroup branches into their respective ka 0.10 branch and undo the new branch redirection before merging this one

maleadt:
> I think the ideal solution would be to have `get_num_sub_groups` always return at least 1 so we can force a sub-group size.

FWIW, I fixed that, and pocl_jll should include the patch merged upstream to make `get_num_sub_groups` behave sensibly.

christiangnrd:
> FWIW, I fixed that, and pocl_jll should include the patch merged upstream to make `get_num_sub_groups` behave sensibly.

I never got around to removing setting the subgroup size by default so no changes necessary.

Also, I bumped the required pocl_standalone_jll to 7.1.3, could/should we yank 7.1.3+0 to guarantee no segfaults from libpocl mismatches?

vchuravy (approved):
Looks good from my side, van all current backends implement this?

christiangnrd:
> Looks good from my side, van all current backends implement this?

I have a branch that passes tests for every backend. I had to implement shuffle_down manually with shuffle for a couple of them.

I guess the newly supported cuda opencl is failing I’ll have to take a look.

src/pocl/compiler/compilation.jl
```diff
@@ -130,14 +139,17 @@ function compiler_config(dev::cl.Device; kwargs...)
     end
     return config
 end
-@noinline function _compiler_config(dev; kernel = true, name = nothing, always_inline = false, kwargs...)
+@noinline function _compiler_config(dev; kernel = true, name = nothing, always_inline = false, sub_group_size::Union{Nothing, Int} = 32, kwargs...)
```
vchuravy, line 142: Did you try different defaults here?

christiangnrd: I didn't. I just went with what is most common on the GPU side

christiangnrd:
@vchuravy I added a `supports_subgroup` function to KA that defaults to true since it only really needs to be defined in OpenCL.jl. If you agree with that approach I'll revert CI pointing to my `intrinsicsnew` branches and merge this!

src/pocl/backend.jl
```diff
@@ -227,10 +229,36 @@ end
 function KI.max_work_group_size(::POCLBackend)::Int
     return Int(device().max_work_group_size)
 end
+function KI.sub_group_size(::POCLBackend)::Int
+    sg_sizes = cl.device().sub_group_sizes
```
vchuravy, line 236: Can you add a comment explaining the logic here?

christiangnrd:
```suggestion
    # POCL can technically support any sub_group size.
    #  Check for common values used on GPUs then
    #  return 1 otherwise
    sg_sizes = cl.device().sub_group_sizes
```

src/KernelAbstractions.jl
```diff
@@ -647,6 +647,17 @@ function ones(backend::Backend, ::Type{T}, dims::Tuple; kwargs...) where {T}
...
+"""
+    supports_subgroups(::Backend)::Bool
+
+Returns whether subgroups are supported by the backend.
+
+!!! note
+    Backend implementations **should** implement this function
+    only if they **do not** support subgroups.
+"""
+supports_subgroups(::Backend) = true
```
christiangnrd: Looking at the subgroup extensions with OpenCL, maybe we need to have a more granular interface to check for subgroups, shuffles, ballots, etc

christiangnrd:
GHA failure unrelated.

PR #679, opened 2026-02-04
Support Julia 1.13

eschnett:


giordano:
Bump 🙂

christiangnrd:
Metal failure is expected, AMDGPU failure is a workspace bug but other versions pass so I'm going to merge this and if anyone objects to the new increased GPU CI load we can reduce it in a separate PR

PR #681, opened 2026-02-06
[nanoPOCL] Re-enable error reporting in 1.12+

christiangnrd:
Partial revert of #674 but keeps the typo fixes


vchuravy:
Blocked by #624 it seems like

vchuravy:
Looks like this was also the Khronos backend.

PR #700, opened 2026-05-28
Use GPUCompiler 1.13.3

vchuravy:
We want to use https://github.com/JuliaGPU/GPUCompiler.jl/pull/812


vchuravy:
@maleadt didn't have time to dig in, but that failure looks new.

maleadt:
Note that this may be a legitimate bug in KA.jl or so, because without the `trap` to `ret` PoCL was aggressive enough to optimize out the print statement too (making it fully invisible in many cases). For example, in OpenCL.jl I had to fix https://github.com/JuliaGPU/OpenCL.jl/pull/436.

maleadt:
Oh my god this is a PoCL miscompilation because of a lane potentially exiting before it hits a barrier. I do have a workaround though... I'm also having Claude investigate a fix in PoCL itself.

maleadt:
So the workaround is benign; @vchuravy please review https://github.com/JuliaGPU/KernelAbstractions.jl/pull/700/changes/f95f72d3954a11972bbfb018a954dedd8875bef5. It essentially avoids some pesky bounds checks when running with `--check-bounds=true`; normal code is unaffected.

I'll fix this in PoCL next.

vchuravy:
Neat! Thank you.

PR #704, opened 2026-06-08
Use pocl_standalone_jll

vchuravy:
- **Use pocl_standalone_jll**
- **fixup! Use pocl_standalone_jll**


.github/workflows/ci.yml
```diff
@@ -103,7 +103,7 @@ jobs:
                          -DKERNELLIB_HOST_CPU_VARIANTS=distro```)
                   run(```$(cmake()) --build $builddir --parallel $(Sys.CPU_THREADS) --target install```)
```
vchuravy, line 119: Need to update the config here to match https://github.com/JuliaPackaging/Yggdrasil/pull/11166 after https://github.com/pocl/pocl/pull/2189 has landed

maleadt: I don't think we need to wait on https://github.com/pocl/pocl/pull/2189 since you don't do a macOS local build here.

I've updated the CI config.

PR #626, opened 2025-07-11
always validate

vchuravy:

PR #706, opened 2026-06-09
Actually use the llvm backend and not the khronos backend

vchuravy:
As noticed by Simeon in https://github.com/JuliaGPU/KernelAbstractions.jl/pull/626#issuecomment-4660666876


christiangnrd:
This could explain why we weren't encountering this issue over in OpenCL with validation on

PR #711, opened 2026-06-12
[CI] Don't precompile needlessly

christiangnrd:
Currently wastes a lot of time precompiling for invalid intermediate env configurations. Disables precompilation at every step of environment preparation to only precompile the actual test environment 

Testing with Metal but if it pans out I'll implement it for all pipelines if it pans out


christiangnrd:
On the CUDA tests, this seems to save ~2-3 minutes per run

PR #714, opened 2026-06-22
[pocl] Back `@private` Scratchpad with GPUCompiler.alloca

vchuravy:
## Summary

Replaces the POCL back-end's `MArray`-backed `@private` scratchpad with a direct per-workitem stack allocation via `GPUCompiler.alloca`. The returned `Ptr` is wrapped in a `CLDeviceArray` over OpenCL "Function" storage (LLVM addrspace 0), which is where the SPIR-V target places allocas.

```julia
@device_override @inline function KA.Scratchpad(ctx, ::Type{T}, ::Val{Dims}) where {T, Dims}
    ptr = POCL.GPUCompiler.alloca(T, Val(prod(Dims)))
    CLDeviceArray(Dims, reinterpret(POCL.LLVMPtr{T, POCL.AS.Function}, ptr))
end
```

This drops the StaticArrays dependency from the POCL back-end (StaticArrays is still used by the CPU back-end).

## Why

`GPUCompiler.alloca` emits a real entry-block `alloca` that the optimizer can promote, in the target's alloca address space — avoiding the unsoundness of `llvmcall` + `alloca` and the overhead/semantics of `MArray`. See the motivation in the companion GPUCompiler PR.

## Alignment

The alloca is aligned to `Base.datatype_alignment(T)`, which is exactly the alignment `CLDeviceArray` uses for its element loads/stores (`alignment(::CLDeviceArray{T})`), so accesses are consistent. `isbits`-union element types are intentionally unsupported (`GPUCompiler.alloca` guards on `isbitstype(T)`).

## Status

**Draft** — depends on **JuliaGPU/GPUCompiler.jl#859** (adds the `alloca` intrinsic). `Project.toml` compat is bumped to `GPUCompiler = "1.23"`; this can be un-drafted once that is merged and released.

## Testing

Verified end-to-end against the local GPUCompiler branch: `@private Float32 (4,)` lowers to `alloca [16 x i8], align 4` in addrspace 0 with no surviving `julia.gpu.alloca`, and the kernel runs correctly on the POCL CPU device.


christiangnrd:
Julia 1.11/1.13 on linux arm test failures unrelated

PR #716, opened 2026-06-29
Remove per-launch heap allocation in CPU `__run`

simone-silvestri:
the current `__run` function leads to heap allocation because it boxes the args.
Apparently the problem is with `len` and `rem` that are boxed because of being defined inside an if condition which leads to boxing of the whole args. In Oceananigans, these args are typically very large (we can get easily to MB) which in turn leads to a lot of GC getting called.

MWE:
```julia
julia> ci = only(code_lowered(KernelAbstractions.__run, NTuple{6, Any}));

julia> !any(e -> occursin("Box", string(e)), ci.code)
false
```

This is a backport given that KA main switched to POCL,

cc @giordano

### DIFF
--- a/src/cpu.jl
+++ b/src/cpu.jl
@@ -97,16 +97,12 @@ end
 # Inference barriers
 function __run(obj, ndrange, iterspace, args, dynamic, static_threads)
     N = length(iterspace)
-    Nthreads = Threads.nthreads()
-    if Nthreads == 1
-        len, rem = N, 0
+    nthreads = Threads.nthreads()
+    Nthreads, len, rem = if nthreads == 1
+        1, N, 0
     else
-        len, rem = divrem(N, Nthreads)
-    end
-    # not enough iterations for all the threads?
-    if len == 0
-        Nthreads = N
-        len, rem = 1, 0
+        l, r = divrem(N, nthreads)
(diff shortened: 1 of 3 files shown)


src/cpu.jl
```diff
@@ -97,16 +97,12 @@ end
...
-        len, rem = divrem(N, Nthreads)
-    end
-    # not enough iterations for all the threads?
-    if len == 0
-        Nthreads = N
-        len, rem = 1, 0
+        l, r = divrem(N, nthreads)
+        l == 0 ? (N, 1, 0) : (nthreads, l, r)
     end
     if Nthreads == 1
```
vchuravy, line 111: *sigh* Julia closure analysis fails again.

Can you instead wrap the  `if static_threads` in a `let len=len, rem=rem ... end`

giordano: That'd also work (we tested that before)

vchuravy: It makes clear why it is needed and prevents me refactoring the code and breaking it again.

giordano:
@simone-silvestri you need to merge on top of the remote target branch to get the buildkite pipeline work (that was fixed by #717)

simone-silvestri:
Thanks @vchuravy, can we release a version 0.9.42?

edit: Ah already done! Thanks

PR #719, opened 2026-06-30
feat(intrinsics): add KI.vload / KI.vstore! for wide vector memory operations

shreyas-omkar:
Adds `KernelIntrinsics.vload` and `KernelIntrinsics.vstore!` 

- **CPU (`Array{T}`):** casts `pointer(arr, idx)` to
  `Ptr{NTuple{N,VecElement{T}}}` and calls `unsafe_load`/`unsafe_store!`.
  Julia/LLVM lowers `NTuple{N,VecElement{T}}` to `<N x T>`, emitting one wide
  op (e.g. `movups` / `vmovaps` on x86).

- **GPU:** each backend registers a `@device_override` for its concrete device
  array type that reinterprets the `LLVMPtr` with `N*sizeof(T)` alignment.
  POCL/SPIR-V override is included here (`CLDeviceArray` in
  `src/pocl/backend.jl`); CUDA.jl and AMDGPU.jl can add analogous overrides
  in their own KA extensions.

- **Fallback:** `N == 1`, non-primitive `T`, or any array without a registered
  override falls back to N scalar accesses via a `@generated` function (avoids
  closures, which GPU compilers reject).

## Benchmark (POCLBackend, Float32, cyclops)

| Size | N | Scalar | vload | Speedup |
|------|---|--------|-------|---------|
| 256K (L2) | 4 | 10.59 GB/s | 11.58 GB/s | 1.09× |
| 256K (L2) | 8 | 11.22 GB/s | 13.03 GB/s | 1.16× |
| 4M (L3)   | 2 | 30.12 GB/s | 39.94 GB/s | 1.33× |
| 4M (L3)   | 8 | 37.46 GB/s | 49.57 GB/s | 1.32× |
| 16M (DRAM)| 4 | 29.89 GB/s | 30.00 GB/s | 1.00× |

Speedup is visible in L2/L3-resident workloads where instruction throughput is
the bottleneck; DRAM-bound workloads are flat as expected.

### DIFF
--- a/src/intrinsics.jl
+++ b/src/intrinsics.jl
@@ -255,13 +255,12 @@ end
     @device_override @inline function KI.vload(::Val{N}, arr::MyDeviceArray{T}, idx::Integer) where {N, T}
     ```
 """
+@inline function vload(::Val{N}, arr::Array{T}, idx::Integer) where {N, T}
+    isprimitivetype(T) && N > 1 && return _vload_ptr(Val(N), pointer(arr, idx))
+    _vload_arr(Val(N), arr, idx)
+end
+
 @inline function vload(::Val{N}, arr::AbstractArray{T}, idx::Integer) where {N, T}
-    if isprimitivetype(T) && N > 1
-        p = pointer(arr, idx)
-        if p isa Ptr
-            return _vload_ptr(Val(N), p)
-        end
-    end
     _vload_arr(Val(N), arr, idx)
 end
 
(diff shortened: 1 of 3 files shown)


simeonschaub:
Did you see https://github.com/eschnett/SIMD.jl/pull/147? That should already provide functionality like `vload`, `vgather` and `vstore` and `vscatter` through SIMD.jl

christiangnrd:
@shreyas-omkar You're using SIMD.jl directly now aren't you? If so can this be closed?

shreyas-omkar:
> @shreyas-omkar You're using SIMD.jl directly now aren't you? If so can this be closed?

Yes. We are using SIMD.jl as per @simeonschaub recommendation. Closing this.

PR #726, opened 2026-07-05
Add Atomix and UnsafeAtomics atomics tests

vchuravy:
Adds an `Atomics` testset to the shared testsuite covering the atomics functionality KernelAbstractions re-exports from Atomix, plus raw-pointer atomics from UnsafeAtomics.

## Atomix (all backends, gated on `supports_atomics(backend())`)
- contended histogram `@atomic +=` for `Int32`/`UInt32`/`Float32`/`Float64`
- `@atomic max`/`@atomic min` reductions across the ndrange
- atomic load (`@atomic A[i]`) and store (`@atomic A[i] = v`)
- `@atomicswap`
- `@atomicreplace` with both succeeding and failing CAS

## UnsafeAtomics (CPU backend only, `@test_skip` elsewhere)
- contended `add!` histogram for `Int32`/`UInt32`/`Float32`/`Float64`
- `max!`/`min!`
- `store!`/`modify!`/`cas!`/`xchg!`/`load` sequence
- `add!` with explicit `seq_cst` ordering

Notes:
- Builds on #737 (merged), which permits the `SPV_EXT_shader_atomic_float_{add,min_max}` extensions based on `cl_ext_float_atomics`, and extends the `test/atomics.jl` it introduced: add and min/max run for `Float32` and, where `supports_float64`, `Float64` through Atomix.
- UnsafeAtomics is added as a direct test dependency, so backend packages that include this testsuite need it in their test environments once they pick up this change. Backends can also opt out via `skip_tests = Set(["Atomics"])`.

### CI status

- Requires **Atomix 1.2.1** (JuliaConcurrent/Atomix.jl#80 and #81, both released) on the GPU backends: the extensions previously implemented `get`/`set!` as `error("not implemented")` and had no swap. Rebased onto `main`, which includes #737 and the `Atomix = "1.2.1"` compat.
- The raw UnsafeAtomics tests run with integer element types only: `atomicrmw fadd`/`fmin`/`fmax` need SPIR-V extensions that NVIDIA's OpenCL driver and OpenCL.jl's C backend do not provide, and UnsafeAtomics has no fallback. Atomix covers floats. The syncscope test uses the system scope only, since NVPTX rejects single-thread scoped atomics and fences.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "KernelAbstractions"
 uuid = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
 authors = ["Valentin Churavy <v.churavy@gmail.com> and contributors"]
-version = "0.10.0-dev"
+version = "0.10.1-dev"
 
 [workspace]
 projects = ["test", "lib/KernelInterface", "lib/KernelInterface/test"]
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "KernelAbstractions"
 uuid = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
 authors = ["Valentin Churavy <v.churavy@gmail.com> and contributors"]
-version = "0.10.1-dev"
+version = "0.10.0-dev"
 
 [workspace]
 projects = ["test", "lib/KernelInterface", "lib/KernelInterface/test"]
--- a/test/atomics.jl
+++ b/test/atomics.jl
@@ -1,0 +1,176 @@
+using KernelAbstractions
+using KernelAbstractions: @atomic, @atomicswap, @atomicreplace
+using Test
+
+import UnsafeAtomics
+
+# Atomix based kernels
+
+@kernel function atomix_add!(hist)
+    i = @index(Global, Linear)
+    j = (i - 1) % length(hist) + 1
+    @inbounds @atomic hist[j] += one(eltype(hist))
+end
+
+@kernel function atomix_max!(A)
+    i = @index(Global, Linear)
+    @inbounds @atomic max(A[1], eltype(A)(i))
+end
+
+@kernel function atomix_min!(A)
+    i = @index(Global, Linear)
+    @inbounds @atomic min(A[1], eltype(A)(i))
+end
+
+@kernel function atomix_load_store!(A, B)
+    i = @index(Global, Linear)
+    @inbounds begin
+        v = @atomic B[i]
+        @atomic A[i] = v
+    end
+end
+
+@kernel function atomix_swap!(A, B)
+    i = @index(Global, Linear)
+    @inbounds B[i] = @atomicswap A[i] = eltype(A)(i)
+end
+
+@kernel function atomix_replace!(A, success)
+    i = @index(Global, Linear)
+    T = eltype(A)
+    @inbounds begin
+        # CAS that must succeed, followed by one that must fail
+        (_, ok1) = @atomicreplace A[i] zero(T) => T(i)
+        (_, ok2) = @atomicreplace A[i] zero(T) => T(-1)
+        success[i] = ok1 & !ok2
+    end
+end
+
+# UnsafeAtomics based kernels (CPU only, operating on raw pointers)
+
+@kernel function unsafe_atomics_add!(hist)
+    i = @index(Global, Linear)
+    j = (i - 1) % length(hist) + 1
+    UnsafeAtomics.add!(pointer(hist, j), one(eltype(hist)))
+end
+
+@kernel function unsafe_atomics_minmax!(A)
+    i = @index(Global, Linear)
+    T = eltype(A)
+    UnsafeAtomics.max!(pointer(A, 1), T(i))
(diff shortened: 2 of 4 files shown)


test/atomics.jl
```diff
@@ -0,0 +1,176 @@
...
+            @test Array(A) == 1:256
+            @test all(Array(success))
+        end
+    end
+
+    @testset "UnsafeAtomics" begin
+        if !(backend() isa CPU)
+            @test_skip "UnsafeAtomics tests only run on the CPU backend"
+            return
+        end
```
vchuravy:
```suggestion
```

vchuravy (commented):
Missing:
- fences
- orderings
- scopes

test/atomics.jl
```diff
@@ -0,0 +1,176 @@
...
+function atomics_testsuite(backend, ArrayT)
+    if !KernelAbstractions.supports_atomics(backend())
+        @test_skip "Backend does not support atomics"
+        return
+    end
+
+    @testset "Atomix" begin
+        # Float32 is excluded since atomic float add requires the SPIR-V
+        # extension SPV_EXT_shader_atomic_float_add, unavailable with PoCL.
+        @testset "atomic add ($T)" for T in (Int32, UInt32)
```
vchuravy, line 141: We should implement fallbacks for this https://github.com/JuliaGPU/GPUCompiler.jl/pull/652

simeonschaub:
> Float atomic add is excluded: the CPU backend compiles through PoCL and atomic float add requires the SPV_EXT_shader_atomic_float_add SPIR-V extension, which PoCL does not provide (LLVM ERROR: The atomic float instruction requires the following SPIR-V extension: SPV_EXT_shader_atomic_float_add).

I just tested it, PoCL supports `SPV_EXT_shader_atomic_float_min_max` just fine, the extension just needs to be enabled:

```julia-repl
julia> using OpenCL, KernelAbstractions, SPIRVIntrinsics, pocl_jll

julia> gentype = Float32; as = 1
1

julia> @eval SPIRVIntrinsics @device_function atomic_max!(p::LLVMPtr{$gentype,$as}, val::$gentype) =
           @builtin_ccall("__spirv_AtomicFMaxEXT", $gentype,
                          (LLVMPtr{$gentype,$as}, UInt32, UInt32, $gentype),
                          p, UInt32(atomic_scope),
                          UInt32(atomic_memory_semantics(Val($as))), val)

julia> function atomix_max!(A)
           @inbounds KernelAbstractions.@atomic max(A[1], eltype(A)(get_global_id()))
           return nothing
       end
atomix_max! (generic function with 4 methods)

julia> a = OpenCL.zeros(T)
0-dimensional CLArray{Float32, 0, OpenCL.cl.UnifiedDeviceMemory}:
0.0

julia> @opencl global_size = 1000 extensions = ["SPV_EXT_shader_atomic_float_min_max"] atomix_max!(a)
OpenCL.HostKernel{typeof(atomix_max!), Tuple{CLDeviceArray{Float32, 0, 1}}}(atomix_max!, OpenCL.Kernel("_Z11atomix_max_13CLDeviceArrayI7Float32Li0ELi1EE" nargs=2), false)

julia> a
0-dimensional CLArray{Float32, 0, OpenCL.cl.UnifiedDeviceMemory}:
1000.0
```

Probably makes sense to enable by default for PoCLBackend, or what do you think?

test/atomics.jl
vchuravy: Done in 291db151 — the CPU-only gate is removed, so the UnsafeAtomics tests now run on every backend.

test/atomics.jl
vchuravy: Added a TODO comment referencing JuliaGPU/GPUCompiler.jl#652 next to the Float32 exclusion (291db151).

vchuravy:
Addressed the review in 291db151:
- UnsafeAtomics tests now run on all backends (CPU-only gate removed, per suggestion).
- **Orderings**: Atomix kernel exercising `:release`/`:acquire` load & store, `:monotonic`/`:acquire_release`/`:sequentially_consistent` RMW, and ordered `@atomicswap`; UnsafeAtomics add parametrized over `monotonic`/`acquire`/`release`/`acq_rel`/`seq_cst`, plus acquire/release load & store.
- **Fences**: non-blocking message-passing test — workitem 1 publishes data with `fence(release)` + monotonic flag store, observers check flag with monotonic load + `fence(acquire)`; anyone who saw the flag must see the data.
- **Scopes**: kernel using system-scope (`none`) contended adds and `singlethread`-scoped store/fence/add.

25 tests passing on the CPU backend.

christiangnrd:
Does this close #308?

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "KernelAbstractions"
 uuid = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
 authors = ["Valentin Churavy <v.churavy@gmail.com> and contributors"]
-version = "0.10.0-dev"
+version = "0.10.1-dev"
```
vchuravy:
```suggestion
version = "0.10.0-dev"
```

PR #728, opened 2026-07-10
Switch POCL backend to CompilerCaching.jl

vchuravy:
Adopts the new GPUCompiler 2.0 caching API for the POCL backend, x-ref https://github.com/JuliaGPU/GPUCompiler.jl/pull/794, following the same approach as https://github.com/JuliaGPU/OpenCL.jl/pull/431:

- Compilation artifacts (SPIR-V bytes, entry name, `device_rng` flag) now live in an `OpenCLResults` struct managed by `GPUCompiler.cached_results`, i.e. attached to the `CodeInstance` in Julia's integrated code cache on 1.11+ (persisting through precompilation) instead of the legacy per-context `cached_compilation` dictionaries.
- Linked `cl.Kernel` handles are session-local and kept in a small per-context linear cache on the results struct; they are not populated while precompiling, so no dangling handles end up in package images.
- `GPUCompiler.reset_runtime()` is gone (the runtime library is cached through the same mechanism now).
- Adds a compilation-cache regression test mirroring the one from the OpenCL.jl PR.

GPUCompiler 2.0 is now registered, so the `[sources]` branch pin is dropped and the `compat` bound is `"2"`. `compile_or_lookup` is adapted to the released `cached_results` API, which returns `nothing` until code exists for the job and is re-fetched after compiling — matching the final form merged in JuliaGPU/OpenCL.jl#431.

Closes #729

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,24 +124,24 @@ jobs:
               end'
             echo '[pocl_standalone_jll]' > test/LocalPreferences.toml
             echo 'libpocl_path="${{ github.workspace }}/target/lib/libpocl_standalone.so"' >> test/LocalPreferences.toml
-      - name: "Co-develop Enzyme and KA"
-        run: |
-          julia -e '
-            using Pkg
-            withenv("JULIA_PKG_PRECOMPILE_AUTO" => 0) do
-              Pkg.activate("test")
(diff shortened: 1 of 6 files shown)


christiangnrd:
Rebasing to pick up #731 should fix the buildkite failures

PR #733, opened 2026-07-20
Rework KernelInterface into a sibling package

vchuravy:
Extracts the `KernelInterface` submodule (`src/interface.jl`) into a standalone sibling subpackage under `lib/KernelInterface`, depended on by KernelAbstractions via a path source.

## Motivation

`KernelInterface` (`KI`) is the low-level API that backends implement against. Living inside KernelAbstractions, it couldn't be depended on independently and dragged the full KernelAbstractions/GPUCompiler stack along. Making it a lightweight sibling package lets backends target the interface directly.

## Changes

- **New `lib/KernelInterface/`** subpackage with its own `Project.toml`. It has **no non-stdlib dependencies**.
- **Dropped the GPUCompiler dependency.** The only symbols used were the small macro helpers `split_kwargs`/`assign_args!`, now vendored verbatim into the package.
- **Decoupled from `KernelAbstractions.Backend`.** The single interface stub that referenced it (`shfl_down_types(::Backend) = DataType[]`) now uses an untyped fallback (`shfl_down_types(_)`), consistent with the neighbouring `multiprocessor_count(_) = 0`.
- **KernelAbstractions** now depends on `KernelInterface` via `[sources] = {path = "lib/KernelInterface"}` and keeps re-exporting `KernelInterface`/`KI`, so all existing `KernelAbstractions.KernelInterface` references (POCL backend, tests, examples) work unchanged.

## Testing

- `KernelInterface` precompiles and loads standalone with zero dependencies.
- Vendored `split_kwargs`/`assign_args!` verified to produce identical output on `@kernel`-style inputs.


lib/KernelInterface/Project.toml
```diff
@@ -0,0 +1,7 @@
+name = "KernelInterface"
+uuid = "4ee993da-d684-4d17-a7dd-4e58e78d92bf"
+authors = ["Valentin Churavy <v.churavy@gmail.com> and contributors"]
+version = "0.1.0"
```
christiangnrd, line 4: Should this match the KA version?

christiangnrd:
Is the goal for backends to not share the fragile deps with the backend or for end-user packages to only need to depend on KernelInterface?

lib/KernelInterface/src/KernelInterface.jl
```diff
@@ -12,8 +12,70 @@ like allocating arrays on a backend.
 """
 module KernelInterface
 
-import ..KernelAbstractions: Backend
-import GPUCompiler: split_kwargs, assign_args!
+## macro tools
```
vchuravy: Backend definition needs to be in KernelInterface

PR #734, opened 2026-07-27
Fix line number handling in `@kernel`

### LINKED ISSUE #732: Code coverage tracking of kernels does not work correctly
The first kernel only gets called on GPU (CUDA), the second kernel get called on GPU (CUDA) and CPU. Codecov does not fully pick them up.

Image

vchuravy:
Fixes #732.

`@kernel` spliced `LineNumberNode`s pointing into KernelAbstractions' own sources into the generated code, and dropped some of the user's line information entirely. GPUCompiler emits device coverage by walking the debug info of `:code_coverage_effect` statements at *compile* time (`record_coverage` in `jlgen.jl`), so bad line info directly becomes bad coverage: kernel lines get reported as untracked, and KernelAbstractions' own sources get credited instead.

### Reproduction

`kernels.jl`, run under `julia --code-coverage`, against `CPU()` (POCL):

```julia
@kernel function mul2!(a)
    i = @index(Global)
    x = a[i]
    a[i] = 2 * x
end

@kernel function sync2!(a)
    i = @index(Local)
    lm = @localmem Float64 (8,)
    lm[i] = a[i]
    @synchronize
    a[i] = lm[i] + 1
end
```

Before — five of the kernels' lines are reported as untracked (`-`) even though both kernels were compiled and run:

```
        - @kernel function mul2!(a)
        1     i = @index(Global)
        1     x = a[i]
        1     a[i] = 2 * x
        - end
        -
        - @kernel function sync2!(a)
        1     i = @index(Local)
        -     lm = @localmem Float64 (8,)
        1     lm[i] = a[i]
        -     @synchronize
        -     a[i] = lm[i] + 1
        - end
```

After — every line is accounted for:

```
        2 @kernel function mul2!(a)
        1     i = @index(Global)
        1     x = a[i]
        1     a[i] = 2 * x
        - end
        -
        2 @kernel function sync2!(a)
        1     i = @index(Local)
        1     lm = @localmem Float64 (8,)
        1     lm[i] = a[i]
        1     @synchronize
        1     a[i] = lm[i] + 1
        - end
```

The lost coverage did not vanish, it was charged to `src/macros.jl`. From the same pre-fix run:

```
        3 function emit(loop)
        3     stmts = Any[]
        -
        3     body = Expr(:block, loop.stmts...)
        3     loopexpr = quote
        -         $(loop.allocations...)
        1         if __active_lane__
        2             $(unblock(body))
```

### Causes

- `emit` built the workitem loop with a `quote` block, so every emitted block started with a `macros.jl` line node. It now builds the expression directly.
- `MacroTools.unblock` strips line nodes when it collapses a single-statement block. This silently discarded the line of the first statement after a `@synchronize` — that statement ended up with *no* line information at all, which is why `a[i] = lm[i] + 1` above reads as untracked. Replaced with a variant that only unwraps blocks carrying no line info.
- `split` hoists `@uniform` / `@localmem` / `@private` out of the workitem loop but left their line nodes behind in the loop body. The pending line node is now tracked and moved along with the statement.

Additionally:

- The generated constructor functions are relocated to the `@kernel` call site, so the `@kernel function …` line is attributed to the user's file instead of `macros.jl:39-46`. That is the `-` → `2` change on the `@kernel` lines above.
- `@groupsize`, `@ndrange`, `@localmem`, `@private`, `@synchronize` and `@print` return bare expressions instead of `quote` blocks, so they no longer inject `KernelAbstractions.jl` line nodes into kernel bodies.

### Testing

Two new test files, both run from `runtests.jl` only (they are backend-independent, so there is no reason to make every backend package pay for them):

- `test/linenumbers.jl` asserts that a `@kernel` expansion only ever refers to the file it was written in, and that every source line of the kernel is represented. Covers the plain case, `@synchronize` plus hoisted allocations, `@uniform`/`@private`, the `inbounds=`/`unsafe_indices=` configurations, and the kernel-language macros.
- `test/coverage.jl` is the end-to-end check: it runs the script above in a subprocess under `--code-coverage` and asserts every kernel line is tracked in the resulting LCOV tracefile. Against the pre-fix expansion it fails on exactly the five lines shown above. Takes ~60-75s, since coverage forces the subprocess to recompile rather than use cached native code.

Coverage is written to an LCOV tracefile in a temp dir rather than via `--code-coverage=user`, which would drop a `.cov` file next to every tracked source in both the checkout and the depot, and would perturb the outer report when the suite itself runs under `Pkg.test(coverage=true)`.

Full suite passes (3125 pass, 4 pre-existing broken), Runic clean.

### Known limitation, not addressed here

Device coverage is only recorded for `--code-coverage=user` and `--code-coverage=all`. The `--code-coverage=@<path>` mode records nothing for device code — verified empirically, and it is the mode itself rather than which files are tracked (pointing `@<path>` at a tree containing both the script and KernelAbstractions' sources still records nothing).

That matters because `Pkg.test(coverage=true)` passes `--code-coverage=@$(pkgroot)`, which is what the standard `julia-actions/julia-runtest` CI setup uses. So under a typical Codecov setup this PR upgrades the `@kernel function …` signature line from untracked to tracked, but kernel *body* lines will still be missing. That gap is in GPUCompiler/Julia, not here.

Device coverage also means "this code was compiled", with counts reflecting compilations rather than executions — inherent to GPUCompiler's approach, since device code cannot call into the Julia runtime.

### DIFF
--- a/src/KernelAbstractions.jl
+++ b/src/KernelAbstractions.jl
@@ -66,7 +66,7 @@ synchronize(dev)
 ```
 """
 macro kernel(expr)
-    return __kernel(expr, #=force_inbounds=# false, #=unsafe_indices=# false)
+    return __kernel(expr, __source__, #=force_inbounds=# false, #=unsafe_indices=# false)
 end
 
 """
@@ -88,7 +88,7 @@ This allows for two different configurations:
 """
 macro kernel(ex...)
     if length(ex) == 1
-        return __kernel(ex[1], false, false)
+        return __kernel(ex[1], __source__, false, false)
     else
         unsafe_indices = false
         force_inbounds = false
@@ -112,7 +112,7 @@ macro kernel(ex...)
                 )
             end
         end
-        return __kernel(ex[end], force_inbounds, unsafe_indices)
+        return __kernel(ex[end], __source__, force_inbounds, unsafe_indices)
     end
 end
 
@@ -267,9 +267,7 @@ a tuple corresponding to kernel configuration. In order to get
 the total size you can use `prod(@groupsize())`.
 """
 macro groupsize()
-    return quote
-        $groupsize($(esc(:__ctx__)))
-    end
+    return :($groupsize($(esc(:__ctx__))))
 end
 
 """
@@ -279,9 +277,7 @@ Query the ndrange on the backend. This function returns
 a tuple corresponding to kernel configuration.
 """
 macro ndrange()
-    return quote
-        $size($ndrange($(esc(:__ctx__))))
-    end
+    return :($size($ndrange($(esc(:__ctx__)))))
 end
 
 """
@@ -293,9 +289,7 @@ macro localmem(T, dims)
     # Stay in sync with CUDAnative
     id = gensym("static_shmem")
 
-    return quote
-        $SharedMemory($(esc(T)), Val($(esc(dims))), Val($(QuoteNode(id))))
-    end
+    return :($SharedMemory($(esc(T)), Val($(esc(dims))), Val($(QuoteNode(id)))))
 end
 
 """
@@ -314,9 +308,7 @@ macro private(T, dims)
     if dims isa Integer
         dims = (dims,)
     end
-    return quote
-        $Scratchpad($(esc(:__ctx__)), $(esc(T)), Val($(esc(dims))))
-    end
+    return :($Scratchpad($(esc(:__ctx__)), $(esc(T)), Val($(esc(dims)))))
 end
 
 """
@@ -350,9 +342,7 @@ workgroup.
     `@synchronize()` must be encountered by all workitems of a work-group executing the kernel or by none at all.
 """
 macro synchronize()
-    return quote
-        $__synchronize()
-    end
+    return :($__synchronize())
 end
 
 """
@@ -372,9 +362,7 @@ workgroup. `cond` is not allowed to have any visible sideffects.
     Since v`0.9.34` this version of the macro is deprecated and lowers to `@synchronize()`
 """
 macro synchronize(cond)
-    return quote
-        $__synchronize()
-    end
+    return :($__synchronize())
 end
 
 """
@@ -458,9 +446,7 @@ macro print(items...)
         end
     end
 
-    return quote
-        $__print($(map(esc, args)...))
-    end
+    return :($__print($(map(esc, args)...)))
 end
 
 """
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -9,8 +9,23 @@ function find_return(stmt)
     return result
 end
 
+# `quote` blocks insert `LineNumberNode`s pointing into this file. Rewriting them
+# to the `@kernel` call site keeps coverage and profiling pointed at the user's
+# code instead of at KernelAbstractions internals.
+relocate_lines(expr, source::LineNumberNode) =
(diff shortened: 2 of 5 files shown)


vchuravy:
Rebased onto `main` and fixed the CI failure.

The coverage test passed locally and on the Windows job, which runs `Pkg.test()` without coverage, but failed everywhere `julia-runtest` runs `Pkg.test(coverage=true)`. The cause is that `Base.julia_cmd()` forwards the parent's `--code-coverage` flag, so under CI the subprocess ran with `--code-coverage=@<pkgroot>` inherited from the parent. That is exactly the path-tracking mode in which GPUCompiler records no device coverage (see the "known limitation" above), and the script itself lives outside the package root, so nothing was tracked at all.

The subprocess command now strips the inherited flags and requests `user` mode explicitly. Verified with a `--code-coverage=@<pkgroot>` parent on Julia 1.10 and 1.12, and with a full `Pkg.test(coverage=true)` run on 1.12.

giordano:
This may have broken Oceananigans on GPU, now most kernels fail to compile with errors like https://buildkite.com/clima/oceananigans/builds/33494#01a09aec-7e20-4cfa-acd5-ba89acad5023/L9
```
Error During Test at /var/lib/buildkite-agent/builds/nautilus-19/clima/oceananigans/test/test_poisson_solvers_stretched_grids.jl:43
  Test threw exception
  Expression: stretched_poisson_solver_correct_answer(Float64, arch, topo, 8, 11, faces; stretched_axis)
  Failed to compile PTX code (ptxas exited with code 255)
  Invocation arguments: --generate-line-info --verbose --gpu-name sm_70 --output-file /tmp/jl_JeJa92Xgpy.cubin /tmp/jl_vABwFELCV8.ptx
  ptxas fatal   : Unexpected non-ASCII character encountered on line 6313
  ptxas fatal   : Ptx assembly aborted due to errors
  If you think this is a bug, please file an issue and attach /tmp/jl_vABwFELCV8.ptx
```

***Edit***: uhm, maybe not this PR because https://github.com/CliMA/Oceananigans.jl/pull/5799 is currently checking out 413576822fdafdb5d1097355a93d47a56dc46372, but it's something recent, it was working on 41a7b238193687e6ea36d5f58bff0ee87823ce27: https://github.com/JuliaGPU/KernelAbstractions.jl/compare/41a7b238193687e6ea36d5f58bff0ee87823ce27...413576822fdafdb5d1097355a93d47a56dc46372 (can't really see anything related to debug info in this range though).  ***Edit 2***: maybe the issue is from https://github.com/JuliaGPU/CUDA.jl/pull/2944, I'm checking out the branch not a specific revision (that branch is being rebased, so having fixed revision wouldn't work very well....)

vchuravy:
Do you have the ptx file?

giordano:
I tried to define simple kernels with greek letters sprinkled everywhere (kernel name, variable names), but couldn't reproduce the error locally.  I'll try to reduce the error we're seeing in Oceananigans tests and open an issue, in any case it's most definitely not related to this PR as the failure predates this PR being merged (but it should somehow have to do with debuginfo because `julia -g0` seems to be a workaround)

giordano:
Turned out to be a bug in CUDA.jl 6.4: https://github.com/JuliaGPU/CUDA.jl/issues/3274

Issue #600, opened 2025-05-20
Switch to POCL


vchuravy:
The outstanding 
- [x] POCL@v7 https://github.com/JuliaGPU/KernelAbstractions.jl/pull/599
- [x] printf support #681 https://github.com/JuliaGPU/KernelAbstractions.jl/pull/592
- [x] Switch to pocl_standalone  #704
  - https://github.com/JuliaPackaging/Yggdrasil/pull/11166
  - Renaming issue in build 
- [x] Distentagle SPIRVIntrinsics: Currently we inherit quirks from OpenCL https://github.com/JuliaGPU/KernelAbstractions.jl/blob/main/src/pocl/device/quirks.jl but this uses the same method table as OpenCL
- [x] Test OpenCL and KA together to ensure that we have no hidden conflicts.
   - [x] Downstream CI ala #608  
- [x] #624 
- [x] #641  (#659)
- [x] float atomics
   - [x] https://github.com/JuliaGPU/KernelAbstractions.jl/issues/671
   - [x] Atomix test https://github.com/JuliaGPU/KernelAbstractions.jl/pull/308
   - [x] https://github.com/JuliaGPU/OpenCL.jl/pull/453 
- [x] Float16 support
  - https://github.com/JuliaGPU/KernelAbstractions.jl/pull/609
  - https://github.com/JuliaGPU/OpenCL.jl/issues/312
- [x] Windows compilation failure 

cc: @maleadt

christiangnrd:
Close?

vchuravy:
Yeah I think we are pretty much there

PR #737, opened 2026-07-27
[pocl] Enable float add and min/max atomics via cl_ext_float_atomics

### LINKED ISSUE #671: POCL backend atomix
@SimonDanisch noticed that this requires the atomic extensions x-ref https://github.com/JuliaGPU/OpenCL.jl/pull/410

vchuravy:
Part of #600 (the last open item: float atomics) — fixes #671.

### Problem

Atomix lowers `@atomic A[i] += x` and `@atomic max(A[i], x)` on floats to LLVM `atomicrmw fadd` / `fmin` / `fmax`. The SPIR-V backend refuses to translate a module containing them unless the guarding extension has been permitted, so on the CPU backend every float atomic fails to compile:

```
LLVM ERROR: The atomic float instruction requires the following SPIR-V extension: SPV_EXT_shader_atomic_float_add
LLVM ERROR: The atomic float instruction requires the following SPIR-V extension: SPV_EXT_shader_atomic_float_min_max
```

Enzyme's reverse mode hits the first one too, because it accumulates gradients with atomic fadd.

### Change

- `nanoOpenCL`: add the `cl_ext_float_atomics` constants and expose the per-precision capability bitfields as `dev.single_fp_atomic_capabilities`, `dev.double_fp_atomic_capabilities`, `dev.half_fp_atomic_capabilities` (zero when the extension is absent).
- `default_spirv_extensions(dev)` permits `SPV_EXT_shader_atomic_float_add` and `SPV_EXT_shader_atomic_float_min_max` when the device natively supports the operation for Float32 or Float64, in both global and local memory (same rule as OpenCL.jl's `has_fp_atomics`). Listing an extension only *permits* it, so kernels without float atomics are unaffected. An explicit `extensions=` keyword still wins.
- Testsuite: new `Atomics` entry running `@atomic` add/max/min for `Int32`, `UInt32`, `Float32` and, where `supports_float64`, `Float64` on every backend (skipped when `supports_atomics` is false). `runtests.jl` additionally checks that pocl's CPU device gets both extensions and that the override works.

pocl's CPU device reports native add and min/max atomics for single and double precision (global and local), and none for half.

### Not in this PR

- Devices *without* native float atomics still fail with the same LLVM error, since Atomix has no POCL hook to fall back to a compare-and-swap loop (SPIRVIntrinsics 1.1 ships `atomic_*_fallback!` for that). Not needed for pocl's CPU device.
- #726 adds the broader Atomix/UnsafeAtomics coverage (orderings, fences, swap/replace); it will be rebased on top of this and extend the same `test/atomics.jl`.
- Enzyme reverse mode additionally needs a `SPIRV_LLVM_Backend_jll` newer than 22.1.x (llvm/llvm-project#200065 plus one pending fix).

### CI status

- Float min/max on the GPU backends needs **Atomix 1.2.1** (JuliaConcurrent/Atomix.jl#80 and #81, both released), which adds a compare-and-swap fallback where a backend has no native instruction and fixes the CUDA Float64 add and float swap paths. This PR requires `Atomix = "1.2.1"` and runs the float min/max cases on every backend.

https://claude.ai/code/session_01DnYMbgSmZQr7KtRUwei63a

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -41,7 +41,7 @@ StaticArraysExt = "StaticArrays"
 
 [compat]
 Adapt = "0.4, 1.0, 2.0, 3.0, 4"
-Atomix = "0.1, 1"
+Atomix = "1.2"
 EnzymeCore = "0.7, 0.8.1"
 GPUCompiler = "2"
 InteractiveUtils = "1.6"
--- a/Project.toml
+++ b/Project.toml
@@ -41,7 +41,7 @@ StaticArraysExt = "StaticArrays"
 
 [compat]
 Adapt = "0.4, 1.0, 2.0, 3.0, 4"
-Atomix = "1.2"
+Atomix = "1.2.1"
 EnzymeCore = "0.7, 0.8.1"
 GPUCompiler = "2"
 InteractiveUtils = "1.6"
--- a/docs/src/examples/atomix.md
+++ b/docs/src/examples/atomix.md
@@ -78,6 +78,8 @@ supported on every backend that reports `KernelAbstractions.supports_atomics(bac
 
 | Operation                                      | `Int32`, `UInt32`, `Int64`, `UInt64` | `Float32`, `Float64`[^1] |
 |:-----------------------------------------------|:------------------------------------:|:------------------------:|
+| `@atomic A[i]`, `@atomic A[i] = x`             | ✓                                    | ✓                        |
+| `@atomicswap A[i] = x`                         | ✓                                    | ✓                        |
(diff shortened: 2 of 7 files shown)


simeonschaub:
Have you seen https://github.com/JuliaGPU/OpenCL.jl/pull/453? Might want to copy some of the logic from there. `SPV_EXT_shader_atomic_float_min_max` might be useful too

christiangnrd:
CUDA failure on 1.10/1.11 seem real?

christiangnrd (approved):
LGTM

PR #739, opened 2026-07-28
Tests for host interface functions

christiangnrd:
`supports_atomics` in Metal.jl is broken and I realized we should probably test the interface definitions

All Claude and I reviewed

### DIFF
--- a/src/KernelAbstractions.jl
+++ b/src/KernelAbstractions.jl
@@ -168,7 +168,7 @@ this asynchronous version, which requires additional lifetime guarantees as docu
 function copyto! end
 
 """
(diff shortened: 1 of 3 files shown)


christiangnrd:
Metal failure is fixed by https://github.com/JuliaGPU/Metal.jl/pull/894.

oneAPI and CUDA fail because their `KA.device!` implementations return a device instead of nothing but that's not specified in the interface so probably just remove the tests?

christiangnrd:
Similarly, `priority!` in AMDGPU returns the symbol instead of nothing, but that's not specified in the interface

PR #742, opened 2026-08-18
[KernelInterface] Add option to auto determine launch sizes with provided ndrange

christiangnrd:
Only up to 3 dimensions.

This adds 2 optional keyword arguments to kernel launches. `ndrange` to specify the grid size to split up, and `max_work_group_size` to request a specific work group size. Though it may be lower depending on the kernel itself.

Also moves the KernelInterface testset from KA to KI.


christiangnrd:
Merging to get the 3-day waiting period over with and can keep breaking if needed

PR #746, opened 2026-08-27
More KernelInterface interface tests

christiangnrd:


lib/KernelInterface/test/interface.jl
```diff
@@ -46,6 +46,38 @@ function test_subgroup_kernel(results)
     return
 end
 
+# The interface documents a concrete return type for each device-side function;
+# these kernels record whether the backend honors them.
+const WorkItemNT = @NamedTuple{x::Int, y::Int, z::Int}
```
vchuravy, line 51: Should we define this alias in the package?

christiangnrd: Maybe? However I was starting to think that we should maybe have the `get_X_size` and `get_X_id` intrinsics take an optional type so people can have their kernels use 4-byte indexingwhen they know they won't need more and it might be annoying to do after we've defined this

christiangnrd: I opened https://github.com/JuliaGPU/KernelAbstractions.jl/issues/752 so I can merge this

Issue #631, opened 2025-07-20
API for # of cores/multiprocessors


christiangnrd:
This may not be the best way to approach this, but to improve the heuristic deciding whether to reduce with blocks or with threads I'm thinking there should be a way to expose the number of cores.

See https://github.com/JuliaGPU/CUDA.jl/blob/e561e7a106684f8e4be59cad98a51cc304c671d2/src/mapreduce.jl#L163-L167 and https://github.com/JuliaGPU/Metal.jl/pull/626

I guess we would also need a way to access the max threads per block/group. Maybe we expose an API specifically for reductions that is essentially an interface for what CUDA has defined in `big_mapreduce_threshold`?

@vchuravy @maleadt @anicusan

vchuravy:
I think it does make sense to query the size of maximum number of workgroup size, and how many SM/Core are available.

We just need to be defensive and allow for a system to say "unknown" or "infinity"

christiangnrd:
Are you thinking both a value for "unknown" and another value for "infinite", or one value for both?

Backend state for max threads per group/block. All of the above are queried via the device:

- CUDA returns an `Int32` (`Cint`) see mepreduce.jl for implementation
- Metal returns an `MTLSize`, which is a struct containing 3 `UInt`values typically (1024,1024,1024) when in reality it's 1024 across the 3 dimensions, not >1 billion so maybe it can just be hardcoded as that until such a time when/if it changes
- oneAPI returns an `Int` `maxTotalGroupSize` from`compute_properties`
- AMDGPU returns an `Int32` (`Cint`) through a similar interface to CUDA (`attribute(device, hipDeviceAttributeMaxThreadsPerBlock)`)
- OpenCL `device.max_work_group_size::UInt32`


Backend state for compute unit count:
- CUDA returns an `Int32` (`Cint`) see mepreduce.jl for implementation
- Metal at the moment would have to be done via a kind of hacky solution but I don't think there's a more official way. See https://github.com/JuliaGPU/Metal.jl/pull/626
- oneAPI **Not sure
- AMDGPU returns an `Int32` (`Cint`) through a similar interface to CUDA (`attribute(device, hipDeviceAttributeMultiprocessorCount)`)
- OpenCL `device.max_compute_units::Csize_t`

christiangnrd:
[...]

PR #750, opened 2026-09-07
Add KernelAbstractions.@spawn and record_event/wait_event

vchuravy:
## Summary

Adds `KernelAbstractions.@spawn backend expr`, which runs `expr` on a new Julia task while keeping the work queued on `backend` ordered between the two tasks. It encodes the discipline users otherwise have to write by hand:

```julia
synchronize(backend)
Threads.@spawn begin
    ...
    synchronize(backend)
end |> wait
```

The macro's protocol:

1. The spawning task calls `record_event(backend)`.
2. The new task selects the spawning task's device with `device!`, then calls `wait_event(backend, event)`.
3. After `expr` returns, the task calls `synchronize(backend)`, so `wait(task)` / `fetch(task)` imply that all of the task's device work has completed.

An optional first argument is forwarded to `Threads.@spawn` as the threadpool (`@spawn :interactive backend expr`). The macro is not exported, to avoid clashing with `Threads.@spawn`.

### Backend opt-in

Two new optional functions in KernelInterface:

- `record_event(backend)` defaults to a full `synchronize` returning `nothing`, which is always correct.
- `wait_event(::Backend, ::Nothing)` is a no-op.

A backend with task-local streams (CUDA.jl, AMDGPU.jl, ...) can override `record_event` to record an event on the current stream without blocking, and implement `wait_event` for that event type as a stream wait, turning the protocol into

```julia
event = record()
@spawn begin
    wait(event)
    ...
    synchronize()
end |> wait
```

Backends with a single global queue (POCL) need no changes.

### Docs

- `synchronize` docstring now recommends a cooperative, non-blocking implementation, since a blocking one serializes otherwise independent spawned tasks.
- API and KernelInterface pages list the new functions; the implementer notes gain a "Task-local queues and `@spawn`" section; the quickstart's task-programming section now recommends `@spawn`.

## Test plan

- [x] New `Spawn` testsuite entry (`test/spawn.jl`): ordering after the parent's queued work, visibility after `wait`, device pinning, threadpool forms, single evaluation of the backend expression, error propagation, many concurrent tasks. Passes on the CPU backend.
- [x] KernelInterface standalone tests cover the `record_event`/`wait_event` fallbacks.
- [x] Full CPU testsuite: all pass except a pre-existing `Private` error caused by my local GPUCompiler dev checkout lacking `alloca` (unrelated to this change).
- [ ] CI on GPU backends (they use the default synchronize-based fallback until they opt in).

https://claude.ai/code/session_01YQWT6DHjjRoh1YaAsFEUN5


docs/src/implementations.md
```diff
@@ -10,4 +10,33 @@ implement a cooperative wait that will `yield` the current task and return the
...
+computation with MPI, and for [`KernelAbstractions.@spawn`](@ref), whose
+trailing `synchronize` would otherwise stall every task scheduled on the same
+thread instead of letting independent tasks run concurrently.
+
+## Task-local queues and `KernelAbstractions.@spawn`
+
+Backends are free to give each Julia task its own queue (stream), so that kernels
+launched from different tasks can execute concurrently. The price is that work queued
+from two tasks is not ordered with respect to each other, and that a task waiting on
+another task with `wait` learns nothing about the state of that task's queue.
```
vchuravy, line 22:
```suggestion
Backends should give each Julia task its own queue/stream,  so that kernels
launched from different tasks can execute concurrently. This implies that work queued
from two tasks is not ordered with respect to each other.
```

docs/src/implementations.md
```diff
@@ -10,4 +10,33 @@ implement a cooperative wait that will `yield` the current task and return the
...
+
+- Before the new task is created, the spawning task calls
+  [`record_event`](@ref KernelAbstractions.record_event) on the backend. The default
+  implementation is a full [`synchronize`](@ref) returning `nothing`, which is always
+  correct. A backend with task-local queues **may** instead record an event on the
+  current task's queue and return it, so that the spawning task does not have to wait.
+- The new task first selects the spawning task's device with [`device!`](@ref KernelAbstractions.device!),
+  then calls [`wait_event`](@ref KernelAbstractions.wait_event) with the recorded handle.
+  A backend that overrides `record_event` **must** implement `wait_event` for its event
+  type, typically by making the current task's queue wait on the event.
```
vchuravy, line 35: Hm... do we need the `device!`?

docs/src/implementations.md
```diff
@@ -10,4 +10,33 @@ implement a cooperative wait that will `yield` the current task and return the
...
+- After the user's code returns, the new task calls [`synchronize`](@ref), so that
+  `wait(task)` in any other task implies that all work queued by the spawned task has
+  completed.
+
+A backend with a single, global queue needs no changes: the defaults are exactly the
+"synchronize before, synchronize after" discipline users would otherwise write by hand.
```
vchuravy, line 41:
```suggestion
```

docs/src/quickstart.md
```diff
@@ -134,34 +134,34 @@ mymul(A, ones(size(A)))
...
-On GPU backends, [`synchronize`](@ref) is **cooperative** — it yields to the Julia scheduler
-rather than blocking inside a driver call, so other tasks can make progress while a kernel runs.
-See [Notes for backend implementations](@ref implementations_notes) for the contract backend authors must follow.
+Some backends give each Julia task its own queue, so kernels launched from two tasks are not
+ordered with respect to each other, and `wait(task)` on its own says nothing about whether
+the kernels that task launched have finished. Use [`KernelAbstractions.@spawn`](@ref) instead
+of `Threads.@spawn` to launch kernels from a task. It orders the new task's work after the
+work the spawning task has already queued, runs it on the same device, and synchronizes the
+backend before the task finishes, so that `wait(task)` and `fetch(task)` guarantee its
+results are ready:
```
vchuravy, line 146: Needs re-wording

vchuravy:
@christiangnrd we should have this for KI 0.2 as well

christiangnrd:
> @christiangnrd we should have this for KI 0.2 as well

Sure! Is the idea that things you write your code assuming events are supported, and on backends where it isn't it just silently falls back to sequential operation?

vchuravy:
Yeah the sequential ordering is the correct thing, and the events are an optimization.

lib/KernelInterface/src/host.jl
```diff
@@ -32,13 +32,62 @@ end
...
     Backend implementations **must** implement this function.
+
+!!! note "Cooperative synchronization"
+    Backend implementations **should** make `synchronize` cooperative rather than blocking.
+    That is, instead of blocking inside a driver call, it should poll or wait on a
+    completion signal while calling `yield` so that other Julia tasks can run in the
+    meantime. A blocking implementation stalls every task scheduled on the same thread,
+    which defeats overlapping kernels with host work or communication, and makes
+    [`KernelAbstractions.@spawn`](@ref)'s trailing `synchronize` serialize otherwise
+    independent tasks.
```
christiangnrd, line 48: Doesn't the quickstart manual say they have to be cooperative?

christiangnrd (commented):
Claude seems to find it very important that someone reading any part of the documentation know that this defaults to synchronize and that if `record_event` is implemented, `wait_event` must also be implemented.

I think the docs should focus more on desired behaviour, and any implementation recommendations/guidelines should be tucked away and clearly marked as such since realistically it'll be one of us adding backend support and users probably won't care about implementation unless things break

christiangnrd:
It'll probably end up in KI 0.2.0, but if I understand correctly, since the backend implementations are just an optimization (and optional), this can release whenever and backends just set compat to whatever version this is released in once they add support

docs/src/implementations.md
christiangnrd: I think wrong device being selected/synchronized is why the OpenCL CUDA tests fail

christiangnrd: Not sure how to best handle this since we never really dealt with https://github.com/JuliaGPU/KernelAbstractions.jl/issues/631#issuecomment-3135051049

christiangnrd: The lazy solution is to add in the docs that the device interface should be supported to use `@spawn` with a non-default device and disable those tests on the OpenCL CUDA runs

christiangnrd:
I've added support for the device selection interface in KA in my Opencl branches (feel free to review https://github.com/JuliaGPU/OpenCL.jl/pull/474/commits/fe59d682e1c1b99aecf42cd9983580b0b08acef5) so once https://github.com/JuliaGPU/KernelAbstractions.jl/pull/754 is in I'll rebase this and the OpenCL CUDA tests should pass!

docs/src/implementations.md
christiangnrd: We do need the `device!` and I've implemented the interface for OpenCL

christiangnrd:
@vchuravy I just pushed a few commits that address my review comments and also some stuff that Claude found. They're all documentation changes, except for https://github.com/JuliaGPU/KernelAbstractions.jl/pull/750/commits/586d110587e7fb9572f1c29f77d1c667ee422195, which you should review particularly more closely.

It also brought up another issue that I didn't fix since I'm not sure what the best approach would be. Bot explanation related to `@spawn`:
> If `expr` throws, the trailing `synchronize` is skipped, so the task's queued kernels may still be running when `wait(task)` throws. The docstring states the completion guarantee unconditionally. Either document that it only holds when `expr` returns, or wrap the body in `try`/`finally`, which risks a failing synchronize masking the original device error. I lean towards documenting it.

vchuravy:
> hich risks a failing synchronize masking the original device error. I lean towards documenting it.

We could do 

```
try
finally
   sync
end
```

but then what do we do if sync throws as well. I tend to agree the right thing is to say: In case of an exceptional exit the task is not synced.

christiangnrd:
> I tend to agree the right thing is to say: In case of an exceptional exit the task is not synced.

Done

docs/src/quickstart.md
```diff
@@ -134,34 +134,32 @@ mymul(A, ones(size(A)))
...
-On GPU backends, [`synchronize`](@ref) is **cooperative** — it yields to the Julia scheduler
-rather than blocking inside a driver call, so other tasks can make progress while a kernel runs.
-See [Notes for backend implementations](@ref implementations_notes) for the contract backend authors must follow.
+Backends may give each Julia task its own queue, so kernels launched from different tasks
+can run concurrently, but are not ordered with respect to each other. Use
+[`KernelAbstractions.@spawn`](@ref) in place of `Threads.@spawn` to launch kernels from a
+task. It behaves like `Threads.@spawn`, and additionally guarantees that the task runs on
+the same device as the spawning task, that its kernels run after everything the spawning
+task had already queued, and that once `wait(task)` or `fetch(task)` returns, its results
+are ready to use:
```
vchuravy, line 146: So I think the question here is:

```
KA.@spawn begin
      device!(1)
      ...
end
```

Will my switching the device be synchronized with the previous work? Or should we not emit the `device!()` call in `@spawn` and require the backend to add it in `wait_event` if it is necessary?

vchuravy: In particular CUDA supports the other semantics that one might care about:

```
cudaSetDevice(0);
cudaEventRecord(event, stream0);

cudaSetDevice(1);
cudaStreamWaitEvent(stream1, event, 0);
```

To synchronize work on device 1 with work on device 0.
Thus to have users use `device!` we also need to support them to sync between devices if necessay

src/spawn.jl
```diff
@@ -0,0 +1,76 @@
...
+    end
+
+    # The whole expansion is escaped so that `Threads.@spawn` is expanded in the caller's
+    # scope: that is what lets an enclosing `@sync` see the task, and what makes `$x`
+    # interpolation in `expr` work. Our own temporaries are gensyms so they cannot clash
+    # with the user's variables.
+    b, dev, event, result = gensym(:backend), gensym(:dev), gensym(:event), gensym(:result)
+    body = quote
+        $KI.device!($b, $dev)
+        $KI.wait_event($b, $event)
```
vchuravy, line 107: So I think this is semantically wrong since a wait_event is stream/queue ordered. 

- CUDA supports wait_event across devices (so we would want to switch devices first then wait)
- If I want to execute `device!` afterwards I would need to sync as well.

One answer is that we would need to tell users:

If you would like to switch devices you must use `synchronize` before `device!`, but ideally we could support the cross device sync from CUDA.

```
ev = KI.record_event(backend)
device!()
KI.wait_event(backend, event)
```

christiangnrd: Are you saying we want to support switching devices within the macro body?

christiangnrd:
@vchuravy I think the order of operations should be merge this, implement in the backend branches, test functionality, if it's all good we release the KernelInterface part as v0.2.1 (or 0.3 if we end up requiring breaking changes)

docs/src/implementations.md
```diff
@@ -10,7 +10,57 @@ implement a cooperative wait that will `yield` the current task and return the
...
+  device has to be selected first. A backend that overrides `record_event` **must**
+  implement `wait_event` for its event type, typically by making the current task's queue
+  wait on the event.
+- After the user's code returns, the new task calls [`synchronize`](@ref), so that
+  `wait(task)` in any other task implies that all work queued by the spawned task has
+  completed.
+
+A new Julia task does not inherit the device of the task that spawned it: backends keep the
+active device in task-local state, which Julia does not copy into a child task, so the task
+starts on the backend's default device. Backends with more than one device therefore
```
vchuravy:
```suggestion
Backends with more than one device
```

docs/src/implementations.md
```diff
@@ -10,7 +10,57 @@ implement a cooperative wait that will `yield` the current task and return the
...
+**must** implement the device interface ([`device`](@ref KernelAbstractions.device),
+[`ndevices`](@ref KernelAbstractions.ndevices), [`device!`](@ref KernelAbstractions.device!))
+for `@spawn` to run on the right device.
+
+`@spawn backend device=id` records the event on the spawning task's device but waits on
+`id`, so a multi-device backend **must** accept an event recorded on a device other than the
+one active in `wait_event`. CUDA expresses this dependency in the driver
+(`cuStreamWaitEvent` works across devices); a backend whose driver cannot **must** fall back
+to waiting cooperatively, as [`synchronize`](@ref) does.
```
vchuravy, line 52:
```suggestion
`@spawn backend device=id` records the event on the spawning task's device but waits on
`id`, so a multi-device backend **must** accept an event recorded on a device other than the
one active in `wait_event`. A backend whose driver cannot **must** fall back
to waiting cooperatively, as [`synchronize`](@ref) does.
```

docs/src/quickstart.md
```diff
@@ -134,34 +134,73 @@ mymul(A, ones(size(A)))
...
```
vchuravy, line 145:
```suggestion
Backends may give each Julia task its own queue, so kernels launched from different tasks
can run concurrently, but are not ordered with respect to each other. Use
[`KernelAbstractions.@spawn`](@ref) in place of `Threads.@spawn` to launch kernels from a
task. It behaves like `Threads.@spawn`, and additionally guarantee that kernels run after everything the spawning
task had already queued, and that once `wait(task)` or `fetch(task)` returns, the kernel results
are ready to use:
```

docs/src/quickstart.md
```diff
@@ -134,34 +134,73 @@ mymul(A, ones(size(A)))
...
+yields to the Julia scheduler, so other tasks keep making progress while a kernel runs.
+
+### Which device a task runs on
+
+Backends keep the active device in task-local state, and Julia does not copy that state into
+a child task. A task started with plain `Threads.@spawn` therefore runs on the backend's
+*default* device, whichever device the spawning task was using — a silent surprise if the
+arrays it captured live elsewhere. `KernelAbstractions.@spawn` selects the device explicitly:
+by default the one active in the spawning task, or the one named by `device`, a 1-based index
+into `1:ndevices(backend)`:
```
vchuravy:
```suggestion
A task started with plain `Threads.@spawn` runs on an implementation defined device for chosen backend.
`KernelAbstractions.@spawn` selects the device explicitly:
by default the one active in the spawning task, or the one named by `device`, a 1-based index
into `1:ndevices(backend)`:
```

src/spawn.jl
```diff
@@ -0,0 +1,128 @@
...
+    Order later work by waiting on the task, or by spawning again.
+
+!!! note
+    Prefer `device=` over calling [`device!`](@ref KernelAbstractions.device!) inside
+    `expr`. A `device!` in the body carries no ordering of its own, so work queued after it
+    is ordered neither against the spawning task nor against what the body queued before
+    the switch; you would have to bracket it with
+    [`record_event`](@ref KernelAbstractions.record_event) and
+    [`wait_event`](@ref KernelAbstractions.wait_event) yourself.
+
```
vchuravy, line 63:
```suggestion
    [`wait_event`](@ref KernelAbstractions.wait_event) yourself.

!!! note
    If `expr` throws the state of the device and the internal queue is unspecified. 
```

vchuravy:
I think this is ready from my side. I adjusted the language on some of the implementation notes.

PR #751, opened 2026-09-07
Document and require `adapt(::Backend, x)`

### LINKED ISSUE #501: Document/Decide on `adapt(Backend, ...)` API
In https://github.com/JuliaGPU/KernelAbstractions.jl/pull/364 I added `adapt(backend, x)` as a natural extension for adapt, instead of having to use something like `adapt(default_array_type(backend), x)` which we don't currently have.

This was inspired by https://github.com/JuliaGPU/KernelAbstractions.jl/pull/353 and #297 

As noted by @maleadt I didn't document it and it is a rather unofficial API that has lead users astray https://github.com/JuliaGPU/CUDA.jl/issues/2459#issuecomment-2270809520

Ideally it would simply be `adapt(::CUDABackend, x) = adapt(CuArray, x)`. I am not sure why I didn't use that definition before.

vchuravy:
Closes #501.

`adapt(backend, x)` has existed since #364 (KernelAbstractions 0.9.0) as a way to move data to a backend without knowing its array type, but was never documented and remained an unofficial API. The decision in #501 was:

> Yeah we should document it and require its implementation. Ideally with the `adapt(::CUDABackend, x) = adapt(CuArray, x)` definition.

This PR makes `Adapt.adapt_storage(::Backend, x)` a documented, required part of the backend interface.

## Interface addition

- `KernelAbstractions` now documents `adapt(backend::Backend, x)` (`src/KernelAbstractions.jl`, next to the other Adapt integration): it converts `x` such that its array storage lives on `backend`, walking structures with `Adapt.adapt_structure` and handing every leaf to `Adapt.adapt_storage(backend, leaf)`. The docstring states that backends **must** implement `Adapt.adapt_storage(::NewBackend, x)`, and that the recommended definition delegates to the backend's array type:

  ```julia
  Adapt.adapt_storage(::CUDABackend, x) = adapt(CuArray, x)
  ```

  This is a docstring on the signature only; there is deliberately no fallback method, since an erroring `(::Backend, ::AbstractArray)` fallback would be ambiguous with the recommended untyped definition, and an untyped one would break the scalar pass-through that Adapt.jl provides.
- `KernelInterface` stays free of non-stdlib dependencies: the requirement is a KernelAbstractions one, and `host.jl` only carries a pointer comment.
- The CPU/POCL backend implements it in `src/pocl/backend.jl` as `Adapt.adapt_storage(::POCLBackend, x::AbstractArray) = isbits(x) ? x : adapt(Array, x)`, replacing the previous `adapt_storage(::POCLBackend, a::Array) = a`. This makes `adapt(CPU(), x)` bring arrays of any backend back to the host through `convert(Array, x)` (no more need for backends to define `adapt_storage(::KA.CPU, ::MyArray)`), while keeping `isbits` arrays (ranges, view indices) as they are, matching what the GPU array types do.

## Docs

- API page: `adapt(backend, x)` is listed under "Backends and arrays".
- KernelInterface page: "Implementing a backend" lists the `Adapt.adapt_storage` requirement and notes it comes from KernelAbstractions.
- Notes for backend implementations: new "Moving data with `adapt`" section.

No changelog entry, since `docs/src/index.md` has no 0.10 section yet.

## Tests

- `test/test.jl` ("adapt" testset of the shared testsuite): round-trips data between the host and the backend, and checks that structures around the arrays are traversed.

## Downstream

The existing definitions in CUDA.jl, AMDGPU.jl, oneAPI.jl, Metal.jl and OpenCL.jl (`adapt_storage(::XBackend, ::Array)` plus an identity for their own array type) keep working, but each backend should switch to the one-line recommended definition, e.g.

```julia
Adapt.adapt_storage(::CUDABackend, x) = adapt(CuArray, x)
```

https://claude.ai/code/session_01DnYMbgSmZQr7KtRUwei63a

### DIFF
--- a/docs/src/implementations.md
+++ b/docs/src/implementations.md
@@ -26,5 +26,3 @@ to the backend's array type, so that `adapt(backend, x)` and
 Adapt.adapt_storage(::CUDABackend, x) = adapt(CuArray, x)
 ```
 
-Adapt.jl's fallback `adapt_storage` is the identity, so a backend without this
-method silently leaves data where it is.
--- a/docs/src/implementations.md
+++ b/docs/src/implementations.md
@@ -11,3 +11,20 @@ scheduling slice to the Julia runtime.
 
 This is of particular import to allow for overlapping of communication and
 computation with MPI.
+
+## Moving data with `adapt`
+
+KernelAbstractions extends [Adapt.jl](https://github.com/JuliaGPU/Adapt.jl) so
+that [`adapt(backend, x)`](@ref Adapt.adapt_storage(::Backend, ::Any)) moves the
+arrays in `x` to `backend`, without the caller having to know the backend's
+array type. Every backend **must** support this by extending
+`Adapt.adapt_storage` for its backend type. The recommended definition delegates
+to the backend's array type, so that `adapt(backend, x)` and
+`adapt(BackendArray, x)` agree:
+
+```julia
+Adapt.adapt_storage(::CUDABackend, x) = adapt(CuArray, x)
+```
+
+Adapt.jl's fallback `adapt_storage` is the identity, so a backend without this
+method silently leaves data where it is.
--- a/docs/src/kernelinterface.md
+++ b/docs/src/kernelinterface.md
@@ -8,9 +8,10 @@ CurrentModule = KernelInterface
 backends implement, and that `KernelAbstractions` builds its higher-level kernel
 language on top of.
 
-It ships as a standalone package under `lib/KernelInterface` with **no
-dependencies outside the standard library**, so a backend can implement the
-interface without taking on `KernelAbstractions` or its compiler stack:
+It ships as a standalone package under `lib/KernelInterface` whose **only
+dependency outside the standard library is
+[Adapt.jl](https://github.com/JuliaGPU/Adapt.jl)**, so a backend can implement
+the interface without taking on `KernelAbstractions` or its compiler stack:
 
 ```julia
 using KernelInterface
@@ -138,6 +139,16 @@ pagelock!
 unsafe_free!
 ```
 
+### Data movement
+
+`adapt(backend, x)` moves the arrays in `x` to `backend`; it is
+[Adapt.jl](https://github.com/JuliaGPU/Adapt.jl)'s `adapt`, dispatching to an
+`Adapt.adapt_storage` method the backend supplies.
+
+```@docs; canonical=false
+Adapt.adapt_storage(::Backend, ::Any)
+```
+
 ### Execution
 
 ```@docs; canonical=false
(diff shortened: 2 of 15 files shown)


lib/KernelInterface/src/host.jl
```diff
@@ -39,9 +39,9 @@ Synchronize the current backend.
 """
 function synchronize end
 
-# Define:
-#   adapt_storage(::Backend, a::Array) = adapt(BackendArray, a)
-#   adapt_storage(::Backend, a::BackendArray) = a
+# KernelAbstractions additionally requires `Adapt.adapt_storage(::Backend, x)`, so that
+# `adapt(backend, x)` moves data to the backend; see its documentation. KernelInterface
+# itself stays free of that dependency.
```
christiangnrd: We don't need this here if it's in the documentation
```suggestion
```

src/KernelAbstractions.jl
```diff
@@ -433,6 +433,40 @@ end
...
+using Adapt
+x = adapt(CUDABackend(), rand(Float32, 8))  # a CuArray
+y = adapt(CPU(), x)                         # an Array again
+```
+
+`adapt` walks `x` with `Adapt.adapt_structure` — through tuples, named tuples, views and
+the other wrappers Adapt.jl knows about, as well as any struct that extends it (see
+`Adapt.@adapt_structure`) — and passes each leaf to `Adapt.adapt_storage(backend, leaf)`.
+Scalars pass through unchanged.
+
```
christiangnrd:
```suggestion

```

src/pocl/backend.jl
```diff
@@ -77,6 +77,12 @@ end
 
 KI.allocate(::POCLBackend, ::Type{T}, dims::Tuple; unified::Bool = false) where {T} = Array{T}(undef, dims)
 
+# `Array` is the CPU array type, so `adapt(CPU(), x)` moves data into plain `Array`s and
+# brings device arrays back to the host through their `convert(Array, ...)`. Adapt.jl's
+# `Array` rule flattens every `AbstractArray` leaf; `isbits` arrays (ranges, view indices)
+# are already usable on every backend and are kept as they are, like the GPU array types do.
+Adapt.adapt_storage(::POCLBackend, x::AbstractArray) = isbits(x) ? x : Adapt.adapt(Array, x)
```
christiangnrd: I feel like this could be shorter

christiangnrd (approved):
I'm not 100% decided on it living in KA instead of KI, but it's easier to move in the KA->KI direction than the other way around, and I agree that KI being dependency-free is nice, so I think we should go with this 

I left a few comments

src/pocl/backend.jl
```diff
@@ -77,6 +77,12 @@ end
 
 KI.allocate(::POCLBackend, ::Type{T}, dims::Tuple; unified::Bool = false) where {T} = Array{T}(undef, dims)
 
+# `Array` is the CPU array type, so `adapt(CPU(), x)` moves data into plain `Array`s and
+# brings device arrays back to the host through their `convert(Array, ...)`. Adapt.jl's
+# `Array` rule flattens every `AbstractArray` leaf; `isbits` arrays (ranges, view indices)
+# are already usable on every backend and are kept as they are, like the GPU array types do.
```
vchuravy:
```suggestion
#  Adapt.jl's `Array` rule converts every `AbstractArray` leaf; `isbits` arrays (ranges, view indices)
# which we want to keep as they are.
```

docs/src/kernelinterface.md
```diff
@@ -199,17 +199,23 @@ A backend must, at minimum:
    [`unsafe_free!`](@ref) are required; the remaining functions under
    [Host-side API](@ref) have fallbacks that only need overriding when the
    defaults don't apply.
-3. `@device_override` the device-side functions it supports. The indexing
+3. Extend `Adapt.adapt_storage(::NewBackend, x)` so that
+   [`adapt(backend, x)`](@ref Adapt.adapt_storage(::Backend, ::Any)) moves
+   data to the backend, preferably by delegating to its array type:
+   `Adapt.adapt_storage(::NewBackend, x) = adapt(NewArray, x)`. This is
+   required by `KernelAbstractions` rather than by `KernelInterface`, which
+   does not depend on Adapt.jl.
```
vchuravy:
```suggestion
   `Adapt.adapt_storage(::NewBackend, x) = adapt(NewArray, x)`.
```

docs/src/implementations.md
```diff
@@ -11,3 +11,20 @@ scheduling slice to the Julia runtime.
...
+`Adapt.adapt_storage` for its backend type. The recommended definition delegates
+to the backend's array type, so that `adapt(backend, x)` and
+`adapt(BackendArray, x)` agree:
+
+```julia
+Adapt.adapt_storage(::CUDABackend, x) = adapt(CuArray, x)
+```
+
+Adapt.jl's fallback `adapt_storage` is the identity, so a backend without this
+method silently leaves data where it is.
```
vchuravy:
```suggestion
```

PR #754, opened 2026-09-08
[KernelInterface 0.2] Take an element type in the indexing queries

christiangnrd:
Ignore backend tests

This is something that people seem to really want.

Yet undocumented caveat that the eltype should be the same in every query for a kernel.

Is this a decent interface?


christiangnrd:
Added to milestone because we should make a decision before 0.10, not because this needs to be merged

lib/KernelInterface/src/device.jl
```diff
@@ -1,89 +1,97 @@
...
+    @device_override get_global_id(::Type{T})::@NamedTuple{x::T, y::T, z::T} where {T}
     ```
+    The zero-argument form forwards to `get_global_id(Int)`.
 """
-function get_global_id end
+get_global_id() = get_global_id(Int)
 
 """
-    get_local_size()::@NamedTuple{x::Int, y::Int, z::Int}
+    get_local_size([::Type{T}=Type(Int)])::@NamedTuple{x::T, y::T, z::T}
```
vchuravy:
```suggestion
    get_local_size([::Type{T}=Int])::@NamedTuple{x::T, y::T, z::T}
```

vchuravy (approved):

giordano (commented):
Yay!  Should have FileCheck tests to make sure the indexing queries indeed used the desired type?

christiangnrd:
> Should have FileCheck tests to make sure the indexing queries indeed used the desired type? 

I think the current tests are sufficient? I tested with bad implementations and they do fail

christiangnrd:
Merging but I can always add more tests if needed

PR #755, opened 2026-09-08
[KernelInterface] Only query kernel_max_work_group_size if necessary

christiangnrd:


christiangnrd:
CI is backed up so I'm going to spare the the AMD and oneAPI runners from running these jobs and merge right away.

`max_wgs` is only called inside the `if` so I'd be surprised if defining it in the `if` caused any issues

PR #765, opened 2026-09-10
Launch kernels over a list of indices

giordano:
Part 2/2 of upstreaming Oceananigans piracy of KA (part 1 is #764). Ref: #757.  CC @simone-silvestri.

> `ndrange` may now be a device vector of `CartesianIndex{N}` or
> `NTuple{N, <:Integer}` elements. Each work item handles one listed index:
> `@index(Global, Cartesian)` and `@index(Global, NTuple)` return it,
> `@index(Global, Linear)` its position in the vector, `@ndrange()` the
> vector's length. The kernel must have a dynamic `ndrange` and a 1-D
> workgroup size, static or given at launch; the partial last workgroup is
> bounds-checked against the vector's length.
> 
> This is how Oceananigans iterates over the active cells of an immersed
> boundary grid. It currently wraps the kernel function, pirates
> `Base.getproperty` on `Kernel`, `Adapt.adapt_structure` on `NDRange`
> and `CompilerMetadata`, `partition`, `expand`, and adds one
> `@device_override __validindex` per GPU backend, which the compiled CPU
> backend's own `::Any` override shadows.
> 
> Implementation:
> - `IndexMap{N}` wraps the vector and yields `CartesianIndex{N}` on
>   indexing; it is the `mapping` of a 1-D `NDRange` (`MappedNDRange`).
>   `expand` and `linear_index` on a `MappedNDRange` look the index up.
> - `partition` dispatches vectors to `mapped_partition`, which validates
>   the kernel and workgroup size and partitions `(length(map),)`; the
>   dynamic-check flag is computed as for any other ndrange.
> - Index validity moves into the generic `__validindex(ctx, groupidx,
>   idx)`, which dispatches on the iteration space (`I in ndrange` for
>   regular spaces, position `<= length(map)` for index maps). The 1-arg
>   `__validindex(ctx)` is now a plain method built on
>   `KernelInterface.get_group_id`/`get_local_id`, so the POCL override is
>   removed; `__index_Global_Linear` goes through `__global_linear` in the
>   same way. Backends that still override `__validindex(ctx)` keep
>   working for regular ndranges and should forward to the 3-arg method
>   to support index maps.
> - `Adapt.adapt_structure` for `NDRange`, `IndexMap` and
>   `CompilerMetadata` moves the vector to the device with the kernel
>   arguments; `CompilerMetadata` treats a vector `ndrange` as
>   `CartesianIndices((length(v),))`.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -97,4 +97,7 @@ KernelAbstractions.NDIteration.DynamicOffset
 KernelAbstractions.NDIteration.extents
 KernelAbstractions.NDIteration.offsets
 KernelAbstractions.NDIteration.linear_index
+KernelAbstractions.NDIteration.IndexMap
+KernelAbstractions.NDIteration.MappedNDRange
+KernelAbstractions.__validindex
 ```
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -91,4 +91,10 @@ KernelAbstractions.@context
 KernelAbstractions.argconvert
 KernelAbstractions.NDIteration.DynamicSize
 KernelAbstractions.NDIteration.StaticSize
+KernelAbstractions.NDIteration.NDRange
+KernelAbstractions.NDIteration.StaticOffset
+KernelAbstractions.NDIteration.DynamicOffset
+KernelAbstractions.NDIteration.extents
+KernelAbstractions.NDIteration.offsets
+KernelAbstractions.NDIteration.linear_index
 ```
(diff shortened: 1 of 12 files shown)


vchuravy:
> - `ndrange` may be a device vector of `CartesianIndex`/`NTuple` indices, running one work item per
  listed index (`kernel(A, ndrange=active_cells)`).

I think that's a step too far. 

The user could have simply written:

`ndrange=length(active_cells)` and then done. `idx = active_cells(@index Global)`

giordano:
Closing in favour of #772.

PR #766, opened 2026-09-11
Cut allocations when launching CPU kernels

giordano:
Cutting down a warm CPU kernel launch from about 50 allocations to 1.

> **Where the allocations came from** (allocation profiler on a warm two-argument launch, 52 allocations, 2208 bytes)
> 
> - Julia does not specialize on `args...` that a method only splats onward. That applied to the `Kernel{POCLBackend}` call, `call` and `set_args!`, so every kernel argument (`CompilerMetadata`, device arrays, `KernelState`) was boxed on its way to `clSetKernelArg`, and `set_args!` dispatched dynamically over `enumerate(args)`. About 30 allocations.
> - Per launch: two `Vector{Csize_t}` for the work sizes, a `Ref` for each argument passed to `clSetKernelArg`, an empty SVM pointer vector that no adaptor rule ever fills, a `Base.Pairs` from splatting keywords through the generated kernel call, and a `clGetDeviceInfo` query for the device's maximum work dimensions.
> - `clfunction`, on every launch: a compiler config, a method-instance lookup, a `CompilerJob`, GPUCompiler's cache query, `Ref` and `Some` temporaries, plus a `Core.Box` for a variable captured by the `get!` closure, all to return the same `HostKernel` as last time.
> 
> **Result**
> 
> | Case | allocations main → branch | median main → branch |
> |---|---|---|
> | launch, static wg, dyn ndrange | 48 → 1 | 348 → 347 μs |
> | launch, static wg, static ndrange | 39 → 2 | 349 → 347 μs |
> | launch 3-D, dyn ndrange | 48 → 1 | 350 → 348 μs |
> | launch, dyn wg autotune | 58 → 15 | 319 → 349 μs |
> | throughput 1-D linear, 2^23 | 49 → 1 | 1.677 → 1.612 ms |
> | throughput 3-D Cartesian | 49 → 1 | 1.682 → 1.697 ms |
> 
> The remaining allocation is the `Ref` receiving the event handle from `clEnqueueNDRangeKernel`. The autotune path keeps 15 because it queries the kernel's work-group info and partitions twice each launch; that can be cached per kernel if wanted. Launch latency itself did not move: the ~300 μs median for a 16-element kernel is POCL's enqueue and event wait, not Julia-side work, so the gain is GC pressure rather than wall time on this backend.

### DIFF
--- a/src/pocl/compiler/execution.jl
+++ b/src/pocl/compiler/execution.jl
@@ -204,8 +204,9 @@ end
 # `HostKernel{F, tt}` instances keyed by their type
 const _kernel_fastpath = Dict{DataType, ResolvedKernel}()
 
-# Reading a `ScopedValue` allocates; outside of any dynamic scope it holds its default.
-@static if VERSION >= v"1.11"
+# On Julia 1.11 to 1.13 reading a `ScopedValue` allocates; outside of any dynamic scope it
+# holds its default, so the read is skipped there.
+@static if v"1.11" <= VERSION < v"1.14-"
     @inline compile_hook_set() = Core.current_scope() !== nothing && GPUCompiler.compile_hook[] !== nothing
 else
     @inline compile_hook_set() = GPUCompiler.compile_hook[] !== nothing
--- a/src/pocl/compiler/execution.jl
+++ b/src/pocl/compiler/execution.jl
@@ -193,8 +193,36 @@ end
 
 const clfunction_lock = ReentrantLock()
 
+# `HostKernel` with the world age and context it was resolved in; valid as long as no method
+# has been defined since and the context is unchanged.
+struct ResolvedKernel
+    world::UInt
+    context::nanoOpenCL.Context
+    kernel::Any
+end
+
+# `HostKernel{F, tt}` instances keyed by their type
+const _kernel_fastpath = Dict{DataType, ResolvedKernel}()
+
+# Reading a `ScopedValue` allocates; outside of any dynamic scope it holds its default.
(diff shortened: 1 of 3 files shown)


src/pocl/compiler/execution.jl
```diff
@@ -194,8 +193,36 @@ end
...
+
+# `HostKernel{F, tt}` instances keyed by their type
+const _kernel_fastpath = Dict{DataType, ResolvedKernel}()
+
+# Reading a `ScopedValue` allocates; outside of any dynamic scope it holds its default.
+@static if VERSION >= v"1.11"
+    @inline compile_hook_set() = Core.current_scope() !== nothing && GPUCompiler.compile_hook[] !== nothing
+else
+    @inline compile_hook_set() = GPUCompiler.compile_hook[] !== nothing
+end
```
giordano: Uhm, I think this was resolved in some version of Julia, right?  Don't remember which one offhand though, would need to check

giordano: Probably only v1.14: https://github.com/JuliaLang/julia/pull/61053

vchuravy:
Can we split this up? In particular the cache fast path needs a closer look

giordano:
> Can we split this up? In particular the cache fast path needs a closer look

They're separate commits already.  Could open separate PRs (although I can't do stacks, as I can't push here)

vchuravy:
> as I can't push here)

You should now be able to

giordano:
Closing in favour of #768

PR #769, opened 2026-09-11
Resolve compiled CPU kernels through a fast path

giordano:
Part 2 of the redo of #766 in this repo


maleadt (changes requested):
This looks questionable. AFAIU, it adds an ad-hoc caching layer over GPUCompiler, for the sake of saving 2 tiny (and thus fast) allocations? That doesn't seem worth it. It's also wrong: `_kernel_fastpath` is only caching on `HostKernel{F,tt}`, so distinct callable objects with the same type therefore return the first cached `HostKernel`, including its stored `f`.

maleadt:
I don't think this needs a secondary cache, just some tweaks on how things are specialized.

maleadt:
Benchmarks look good.

giordano:
Is this good to go now? 👀

PR #772, opened 2026-09-11
Launch kernels over a list of indices

giordano:
Redo of #765, in this repo.  Ref: #757.  CC @simone-silvestri.


christiangnrd:
Reposting @vchuravy's comment here:

> > - `ndrange` may be a device vector of `CartesianIndex`/`NTuple` indices, running one work item per
  listed index (`kernel(A, ndrange=active_cells)`).
>
> I think that's a step too far. 
>
> The user could have simply written:
>
> `ndrange=length(active_cells)` and then done. `idx = active_cells(@index Global)`

_Originally posted by @vchuravy in https://github.com/JuliaGPU/KernelAbstractions.jl/issues/765#issuecomment-5630222543_

giordano:
Based on quick testing, we may not need this in Oceananigans after all, and #771 should be sufficient (it's hard to do full testing at the moment though, because of https://github.com/JuliaGPU/KernelAbstractions.jl/pull/734#issuecomment-5656893535).  I'm leaving this open in case it's still useful for #779.

Issue #652, opened 2025-11-03
`@Const` can lead to wrong results on CPU in Julia 1.11/1.12


maximilian-gelbrecht:
MWE below @vchuravy 

Works in 1.10, but not in 1.11/1.12 

```Julia
"""
Minimal working example to demonstrate the difference between:
1. Using `array[I] +=` in a loop (BROKEN)
2. Using a local accumulator then `array[I] =` (WORKS)

This appears to be a bug in KernelAbstractions.jl or me misunderstanding how it handles
compound assignment operators on array elements within loops.
"""

using KernelAbstractions
using Test

# Version 1: Using += directly on array element (BROKEN)
@kernel function accumulate_broken!(output, input, @Const(n))
    I = @index(Global, Cartesian)
    i = I[1]
    j = I[2]
    
    # This should accumulate input[i, k] for k in j:n
    # But it produces WRONG results!
    for k in j:n
        output[I] += input[i, k]
    end
end

# Version 2: Using local accumulator (WORKS)
@kernel function accumulate_fixed!(output, input, @Const(n))
    I = @index(Global, Cartesian)
    i = I[1]
    j = I[2]
    
    # Use local accumulator
    sum_val = zero(eltype(output))
    for k in j:n
        sum_val += input[i, k]
    end
    output[I] = sum_val
end

# Test function
function test_kernel_bug()
    println("="^70)
    println("Minimal Kernel Bug Demonstration")
    println("="^70)
    
    # Setup test data
    n = 8
    m = 5
    
    input = Float32[i + k for i in 1:m, k in 1:n]
    output_broken = zeros(Float32, m, n)
    output_fixed = zeros(Float32, m, n)
    output_cpu = zeros(Float32, m, n)
    
    println("\nInput matrix ($(m)×$(n)):")
    display(input)
    println("\n")
    
    # CPU reference implementation
    println("Computing CPU reference...")
    for i in 1:m
        for j in 1:n
            for k in j:n
                output_cpu[i, j] += input[i, k]
            end
        end
    end
    
    # Version 1: Broken (using +=)
    println("Running BROKEN kernel (using output[I] += ...)...")
    backend = CPU()
    kernel_broken! = accumulate_broken!(backend)
    kernel_broken!(output_broken, input, n, ndrange=size(output_broken))
    KernelAbstractions.synchronize(backend)
    
    # Version 2: Fixed (using local accumulator)
    println("Running FIXED kernel (using local accumulator)...")
[...]

PR #776, opened 2026-09-12
Do not emit aliasscope markers on Julia 1.11 and later

vchuravy:
Julia miscompiles code placed inside an `aliasscope`, see https://github.com/JuliaLang/julia/issues/63129 (and our earlier report in #652). We already restricted the markers to kernels with `@Const` arguments; this disables them entirely on 1.11 and later.

`@Const` arguments are still wrapped in `Base.Experimental.Const` (so back-ends that specialize on it, e.g. for `__ldg`, are unaffected) — they just no longer participate in alias-scope based optimizations on affected versions.

Tested locally on 1.12.7: the `@kernel` expansion of a kernel with a `@Const` argument no longer contains `aliasscope`, and the CPU test suite passes (the Enzyme extension tests error only because Enzyme isn't installed in my local test env).

Note: this does not bump the version in `Project.toml`.

https://claude.ai/code/session_01Be2cuG5qj1BfmHFnsKADNZ

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -78,6 +78,12 @@ function __kernel(expr, generate_cpu = true, force_inbounds = false, unsafe_indi
     end
 end
 
+# Julia 1.11 and later miscompile code placed inside an `aliasscope`, even when the
+# kernel makes no use of `@Const`, so the markers are only emitted on older versions.
+# See https://github.com/JuliaGPU/KernelAbstractions.jl/issues/652
+# and https://github.com/JuliaLang/julia/issues/63129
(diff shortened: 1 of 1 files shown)


stillyslalom:
Potentially fixed upstream by https://github.com/JuliaLang/julia/pull/63130, but this PR should help for v1.11-1.13 if the upstream fix isn't backported.

christiangnrd:
Non-lts releases that aren’t current typically don’t get releases. That convention was broken for 1.11 a few times but I wouldn’t expect it

PR #779, opened 2026-09-13
Add `foreach_index`

vchuravy:
Adds `foreach_index(f, itr, backend = get_backend(itr); workgroupsize = nothing)`, which runs `f`
once per index of `eachindex(itr)` with one work item per index, so a loop over the indices of an
array needs no `@kernel` definition:

```julia
function scale!(y, x)
    foreach_index(x) do i
        @inbounds y[i] = 2 * x[i] + 1
    end
    return y
end
```

`f` receives the index a `for i in eachindex(itr)` loop would produce: a linear index for an array
with `IndexLinear` style, a `CartesianIndex` otherwise.

Ported from [AcceleratedKernels.jl](https://github.com/JuliaGPU/AcceleratedKernels.jl), where
`foreachindex` is the primitive that `map`, `reduce`, `sort`, `accumulate`, `findall`,
`searchsorted` and `reverse` are built on, so it has had a lot of use.

### Based on `mg/ndrange-index-map`

Based on `mg/ndrange-index-map` rather than `main`, because that branch is what makes this a thin
wrapper: the index space is carried by the `ndrange`, so

- the kernel takes no argument besides the function, where AcceleratedKernels passes
  `eachindex(itr)` as a kernel argument and indexes it with the global index, and
- an index space that does not start at 1 needs no special handling — `@index(Global, Cartesian)`
  already returns the shifted index. Covered by a test with an `IndexLinear` array whose `axes1`
  is `-2:7`.

`1f7ed89a` ("Compute the global linear index arithmetically") also removes the `LinearIndices`
lookup that pushed the AcceleratedKernels version to `unsafe_indices=true`, so this uses the
ordinary bounds-checked path.

### What the port drops

The AcceleratedKernels signature carries a CPU scheduling policy (`max_tasks`, `min_elems`,
`prefer_threads`) that selects a `Threads.@spawn` loop over the kernel. 0.10 has no host-threaded
execution path to select — `CPU === POCLBackend` and `get_backend(::Array) === POCLBackend()` —
and that policy is a backend's business anyway, which is where it lived in 0.9 (`CPU_GRAINSIZE`
and the `static` flag). `block_size=256` becomes `workgroupsize`, KA's name for it, and defaults
to `nothing` so that the backend's autotuning picks it.

### Open questions

- **The name.** Spelled with an underscore so that it does not collide with the
  AcceleratedKernels export, and so that the difference in behaviour (no scheduler keywords,
  asynchronous, no `@inbounds` elision) does not hide behind an identical name. The cost is that
  it no longer mirrors `eachindex`, and that the ecosystem carries two near-identical names. The
  alternative would be `foreachindex` left unexported.
- **Compile latency on `CPU()`.** Every distinct closure is a fresh POCL compile, so a one-shot
  `foreach_index` over an `Array` is far more expensive than the threaded loop it reads like. Fine
  for a loop that runs more than once, but worth saying out loud before users reach for this as a
  `@threads` replacement.
- **`foraxes`.** Not included: with an `ndrange` that takes ranges, `foreach_index(f, axes(A, 2))`
  covers it (with the backend passed explicitly, since a range carries no backend).
- **Asynchronous.** This follows every other launch in KA and does not synchronize, even though it
  reads like a `for` loop that has finished. Documented, but it is a plausible footgun.

### Testing

`test/foreach_index.jl`, wired into the test suite: linear index spaces in 1-3 dimensions (result
plus each index visited exactly once), a `CartesianIndices` space through a strided view, an
offset linear space, an index space with no device memory (`1:100`), an explicit `workgroupsize`,
an explicit backend, an empty index space, and the error for an index space that is neither.
Passes on the CPU backend; **not yet run on a GPU backend.**


giordano:
I rebased this on top of #772, previous rebase looked broken.  In case you wanted to go back to previous state, old commit was c033aef7c7127860a78d0a1b52e6d7e1751b9237

src/foreach_index.jl
```diff
@@ -0,0 +1,75 @@
...
+# function: `@index(Global, Cartesian)` already returns the index of the `ndrange`, including
+# the offset of an index space that does not start at 1.
+#
+# `f` is inlined on purpose: an out-of-line call to a closure can spill its captures to local
+# memory on some backends.
+
+@kernel function foreach_index_linear_kernel(f)
+    I = @index(Global, Cartesian)
+    @inline f(I.I[1])
+end
```
vchuravy, line 13:
```suggestion
@kernel function foreach_index_linear_kernel(f)
    I = @index(Global, Linear)
    @inline f(I)
end
```

PR #781, opened 2026-09-14
Document the contract of custom `NDRange` mappings

giordano:


docs/src/implementations.md
```diff
@@ -26,3 +26,25 @@ to the backend's array type, so that `adapt(backend, x)` and
...
+- [`expand(iterspace, groupidx, idx)`](@ref KernelAbstractions.NDIteration.expand)
+  gives the `CartesianIndex` handled by work item `idx` of workgroup `groupidx`;
+- `expand(iterspace, groupidx, idx) in __ndrange(ctx)` tells whether that work
+  item has an index to handle, which is how `__validindex` must be implemented;
+- [`linear_index(__ndrange(ctx), I)`](@ref KernelAbstractions.NDIteration.linear_index)
+  gives the linear index of `I`.
+
+A backend must not assume that `__ndrange(ctx)` is a `CartesianIndices` or that
+`expand` is an affine map: the `mapping` field of the `NDRange` lets a package
+define its own iteration space, for example a list of indices to visit, by
```
vchuravy, line 45: This is a new contract, I am not supper happy about. These are private implementation details and a packing extending them is doing fishy work, and potentially type piracy. 

I would be much happier if we had a public interface that folks could extend

```
abstract type Mapping end
```

And then Oceananigans could do its subtype of that, and so `ndrange=Mapping()`

PR #783, opened 2026-09-14
Check generated device code with FileCheck.jl

vchuravy:
The test suite covers what kernels compute, but nothing pins down *how* they compile. Several of the guarantees KernelAbstractions makes are about the generated code rather than the result, and all of them can regress silently today:

- the global index is computed from the SPIR-V work-item builtins, with no call back into the Julia runtime, and array accesses go to address space 1;
- without `@inbounds` the kernel carries the out-of-bounds path (the error `printf` plus `gpu_report_exception`/`gpu_signal_exception`), and `@inbounds` removes all of it;
- with both the workgroupsize and the ndrange known at compile time, `__validindex` folds away and the kernel body becomes straight-line code (paired with a dynamic-ndrange kernel that still branches, so the check is not vacuous);
- `@localmem` becomes a module-level allocation in address space 3 that the kernel reads and writes directly;
- `@synchronize` becomes a SPIR-V control barrier;
- `@atomic` becomes a native `atomicrmw`, not a compare-and-swap loop;
- `@print` becomes a single variadic `printf` call, not one call per argument.

`test/codegen_checks.jl` states each of those as LLVM FileCheck directives over `@device_code_llvm` output, using [FileCheck.jl](https://github.com/JuliaLLVM/FileCheck.jl). `test/codegen.jl` runs it.

### Why a subprocess

`Pkg.test` defaults to `--check-bounds=yes`, which forces bounds checks on regardless of `@inbounds`. Under it the two `@inbounds` assertions are not merely untestable but false. Making them conditional would mean they never run in CI, so instead `test/codegen.jl` runs the checks in a subprocess that inherits neither `--check-bounds` nor `--code-coverage`, following the existing pattern in `test/coverage.jl` — coverage instruments the very kernels being inspected, and under `Pkg.test(coverage=true)` would also fold the subprocess into the outer report. `codegen_checks.jl` refuses to run unless bounds checking is on `auto`, so a future regression in the flag handling fails loudly rather than quietly asserting nothing.

To iterate on a pattern, run it directly:

```
julia --project=test test/codegen_checks.jl
```

### Notes

- The tests deliberately live outside `Testsuite`. That module is included by the backend packages, which cannot pick up new test dependencies, and the patterns are specific to the in-tree POCL/SPIR-V back-end anyway.
- Patterns avoid pointer syntax, so they hold for both the typed pointers of LLVM 15 (Julia 1.10) and the opaque pointers of later versions. Confirmed on the 1.10 jobs.
- Exclusions use `implicit_check_not` rather than a trailing `@check_not`: a `CHECK-NOT` after the last positive check only covers the range from that match to the end of the input, which for `ret void` is nothing at all.
- The two `@localmem` accesses are `@check_dag`, because LLVM is free to emit the basic blocks in any order — it does in fact sink the barrier block above them here.
- `FileCheck` v1.2 depends only on `LLVM_utils_unified_jll`, a FileCheck-only artifact, rather than on a full LLVM build.

### Testing

`Pkg.test(coverage=true)` passes locally, and CI is green across the matrix.

Each check was verified non-vacuous by mutating the kernel it covers — dropping `@inbounds`, `@synchronize` and `@atomic` each fail exactly the corresponding test. Breaking a pattern was also verified to fail the parent testset, with the subprocess's annotated FileCheck output surfaced in the log.


giordano:
Also https://github.com/JuliaGPU/KernelAbstractions.jl/pull/754#pullrequestreview-5184025394?

PR #784, opened 2026-09-14
Replace `@ka_code_llvm` with GPUCompiler's `@device_code_*`

vchuravy:
`@ka_code_llvm` is dead code on 0.10-dev.

It reflected on the host-side lowering of a kernel and explicitly rejected GPU backends:

```julia
if isa($kern, Kernel{G} where {G <: GPU})
    error("@ka_code_llvm does not support GPU kernels")
end
```

Since the CPU backend became an OpenCL backend and `const CPU = POCLBackend <: GPU`, that guard fires for *every* kernel on *every* backend, so the macro cannot be called at all:

```
julia> @ka_code_llvm mul2(CPU(), 16)(A, ndrange = 64)
ERROR: @ka_code_llvm does not support GPU kernels
```

Nothing in the test suite covered it, which is why it went unnoticed.

## What this does

Removes `@ka_code_llvm`/`ka_code_llvm`, and forwards GPUCompiler's `@device_code_*` macros from `KernelAbstractions` in its place:

```julia
KernelAbstractions.@device_code_llvm debuginfo=:none mul2(backend, 64)(A, ndrange=length(A))
```

These install a compilation hook for the duration of the wrapped expression, so they report on the code a backend actually generates rather than on the host-side lowering, and they do so for any GPUCompiler-based backend — the in-tree CPU backend as well as CUDA, AMDGPU, oneAPI or Metal. That makes them a strictly more capable replacement than the CPU-only macro they replace. The one behavioural difference worth knowing, documented in the docstrings: the wrapped expression is really evaluated, so the kernels it launches are compiled *and* run.

They are **not exported**, because the backend packages export macros of the same name and `using CUDA, KernelAbstractions` would otherwise make the name ambiguous; hence the qualified spelling above.

`@ka_code_typed` is untouched — it reflects on the kernel function as the host sees it, which is backend-independent and still works. Its docstring now points at the `@device_code_*` macros for looking at generated device code.

Also: a `Reflection` section rewrite in `docs/src/api.md`, a 0.10 changelog entry, and a test in `test/runtests.jl` asserting that `KernelAbstractions.@device_code_llvm` over a kernel launch produces IR for that kernel and runs it.

## Testing

- `julia --project=test test/runtests.jl`: 2484 pass, 2 errors — both the pre-existing `ifelse on aggregate types` SPIR-V translation failures (`ComplexF32`/`ComplexF64`), which reproduce unchanged on `main`.
- Docs build clean: no missing-docs and no broken cross-reference warnings, all six macros render on the API page.
- Aqua and Runic clean.


docs/src/api.md
```diff
@@ -72,12 +72,27 @@ KernelAbstractions.backend
...
-These macros help inspect the generated kernel code. LLVM IR reflection via
-[`@ka_code_llvm`](@ref) is only supported on the CPU backend.
+[`@ka_code_typed`](@ref) inspects the kernel function as the host sees it, before the
+backend has compiled it.
+
+To look at the code a backend actually generates, wrap a kernel launch in one of the
+`@device_code_*` macros below. They are GPUCompiler's, and report on every kernel compiled
+while the wrapped expression runs, so they work the same on the CPU backend and on GPU
+backends. They are not exported, because the backend packages (CUDA.jl, AMDGPU.jl, ...)
+export macros of the same name; call them qualified:
```
christiangnrd:
```suggestion
To look at the code a backend actually generates, wrap a kernel launch in one of the
`@device_code_*` macros below. They work the same on the CPU backend and on GPU
backends, and they are public, but not exported, so you must call them qualified:
```

(and then make them public on 1.11+)

docs/src/api.md
```diff
@@ -72,12 +72,27 @@ KernelAbstractions.backend
...
+
+```julia
+KernelAbstractions.@device_code_llvm mul2(backend, 64)(A, ndrange=length(A))
+```
 
 ```@docs
 @ka_code_typed
```
vchuravy: Should also be removed

PR #787, opened 2026-09-18
Fix broken AQUA test

christiangnrd:
Broken by I assume [AQUA.jl v0.8.17](https://github.com/JuliaTesting/Aqua.jl/releases/tag/v0.8.17)

Also include some NFC Project.toml moving around to save on the hassle of having to revert it every Pkg operation


christiangnrd:
@vchuravy I'm going to merge this to get CI green again and we can address the issue (if necessary) in a follow-up PR

